PR #16, opened 2019-07-06
Some package rearranging, Webhook CRD

wbrefvem:


rawlingsj:
@wbrefvem it looks like you’ve merged this before CI validated the PR, also not sure why you’ve moved the charts under /deployments which is non standard for Jenkins X. I think we need to revert this and please ensure CI passes and preview environment works before merging.

Issue #77, opened 2019-08-05
Cleanup the prototype 


ccojocar:
Cleanup the prototype and review the changes done to get the prow's hook working with git-scm. This does not include any refactoring related to jx interface. This is going to be address by #61.

PR #90, opened 2019-08-05
fix: refactor PlumberJob

jstrachan:
* remove unnecessary data and Status
* renamed to avoid confusion that its a CRD - a Plumber implementation can make any kind of CRD it wants

#77

### DIFF
--- a/pkg/plumber/constants.go
+++ b/pkg/plumber/constants.go
@@ -8,10 +8,10 @@ const (
 	// that the pod is running.
 	PlumberJobTypeLabel = "lighthouse.jenkins-x.io/type"
 	// PlumberJobIDLabel is added in resources created by lighthouse and
-	// carries the ID of the PlumberJob that the pod is fulfilling.
-	// We also name resources after the PlumberJob that spawned them but
+	// carries the ID of the PlumberArguments that the pod is fulfilling.
+	// We also name resources after the PlumberArguments that spawned them but
 	// this allows for multiple resources to be linked to one
-	// PlumberJob.
+	// PlumberArguments.
 	PlumberJobIDLabel = "lighthouse.jenkins-x.io/id"
 	// PlumberJobAnnotation is added in resources created by lighthouse and
 	// carries the name of the job that the pod is running. Since
--- a/pkg/plumber/constants.go
+++ b/pkg/plumber/constants.go
@@ -8,10 +8,10 @@ const (
 	// that the pod is running.
 	PlumberJobTypeLabel = "lighthouse.jenkins-x.io/type"
 	// PlumberJobIDLabel is added in resources created by lighthouse and
-	// carries the ID of the PlumberArguments that the pod is fulfilling.
-	// We also name resources after the PlumberArguments that spawned them but
+	// carries the ID of the PipelineOptions that the pod is fulfilling.
+	// We also name resources after the PipelineOptions that spawned them but
 	// this allows for multiple resources to be linked to one
-	// PlumberArguments.
+	// PipelineOptions.
 	PlumberJobIDLabel = "lighthouse.jenkins-x.io/id"
 	// PlumberJobAnnotation is added in resources created by lighthouse and
 	// carries the name of the job that the pod is running. Since
--- a/pkg/plumber/types.go
+++ b/pkg/plumber/types.go
@@ -7,34 +7,34 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
-// PlumberJobType specifies how the job is triggered.
-type PlumberJobType string
+// PipelineKind specifies how the job is triggered.
+type PipelineKind string
 
 // Various job types.
 const (
 	// PresubmitJob means it runs on unmerged PRs.
-	PresubmitJob PlumberJobType = "presubmit"
+	PresubmitJob PipelineKind = "presubmit"
 	// PostsubmitJob means it runs on each new commit.
-	PostsubmitJob PlumberJobType = "postsubmit"
+	PostsubmitJob PipelineKind = "postsubmit"
 	// Periodic job means it runs on a time-basis, unrelated to git changes.
-	PeriodicJob PlumberJobType = "periodic"
+	PeriodicJob PipelineKind = "periodic"
 	// BatchJob tests multiple unmerged PRs at the same time.
-	BatchJob PlumberJobType = "batch"
+	BatchJob PipelineKind = "batch"
 )
 
-// PlumberArguments contains the arguments to pass to the Plumber to create a Tekton Pipeline
-type PlumberArguments struct {
+// PipelineOptions contains the arguments to pass to the Plumber to create a Tekton Pipeline
+type PipelineOptions struct {
 	metav1.TypeMeta   `json:",inline"`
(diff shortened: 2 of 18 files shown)


pkg/plumber/constants.go
```diff
@@ -8,10 +8,10 @@ const (
 	// that the pod is running.
 	PlumberJobTypeLabel = "lighthouse.jenkins-x.io/type"
```
ccojocar: This seems to be named `PlumberJob`

pkg/plumber/constants.go
```diff
@@ -8,10 +8,10 @@ const (
...
 	// PlumberJobIDLabel is added in resources created by lighthouse and
-	// carries the ID of the PlumberJob that the pod is fulfilling.
-	// We also name resources after the PlumberJob that spawned them but
+	// carries the ID of the PlumberArguments that the pod is fulfilling.
+	// We also name resources after the PlumberArguments that spawned them but
 	// this allows for multiple resources to be linked to one
-	// PlumberJob.
+	// PlumberArguments.
```
ccojocar: Why not `PipelineOptions` instead of `PlumberArguments`? It sounds to me more appropriate for this.

pkg/plumber/types.go
```diff
@@ -22,48 +22,19 @@ const (
...
-// By default we tend to turn Webhoks into tekton Pipeline CRDs
-type PlumberJob struct {
+// PlumberArguments contains the arguments to pass to the Plumber to create a Tekton Pipeline
+type PlumberArguments struct {
 	metav1.TypeMeta   `json:",inline"`
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
-	Spec   PlumberJobSpec   `json:"spec,omitempty"`
-	Status PlumberJobStatus `json:"status,omitempty"`
+	Spec PlumberJobSpec `json:"spec,omitempty"`
```
ccojocar: `PlumberJobSpec` doesn't sounds consistent with `PlumberArguments`, maybe we should rename it.

pkg/plumber/types.go
```diff
@@ -22,48 +22,19 @@ const (
...
 }
 
 // PlumberJobSpec the spec of a pipeline request
 type PlumberJobSpec struct {
```
ccojocar: Still `job` in the name?

pkg/plumber/types.go
```diff
@@ -22,48 +22,19 @@ const (
...
 	// Type is the type of job and informs how
 	// the jobs is triggered
 	Type PlumberJobType `json:"type,omitempty"`
```
ccojocar: not sure about the naming `PlumberJobType` => `PlumberType`

pkg/plumber/types.go
```diff
@@ -74,9 +45,6 @@ type PlumberJobSpec struct {
 	// ExtraRefs are auxiliary repositories that
 	// need to be cloned, determined from config
 	ExtraRefs []Refs `json:"extra_refs,omitempty"`
```
ccojocar: Where are we using this `ExtraRefs` argument?

pkg/plumber/types.go
```diff
@@ -86,41 +54,6 @@ type PlumberJobSpec struct {
 	// MaxConcurrency restricts the total number of instances
 	// of this job that can run in parallel at once
 	MaxConcurrency int `json:"max_concurrency,omitempty"`
```
ccojocar: `MaxConcurrency` where is this used? This is specific to prow's controllers. Why do you still want to keep this?

pkg/prow/config/config.go
```diff
@@ -139,14 +139,14 @@ type Controller struct {
 	// JobURLTemplateString compiles into JobURLTemplate at load time.
 	JobURLTemplateString string `json:"job_url_template,omitempty"`
```
ccojocar: It seems confusing, there is still `jobs` in the names.

pkg/prow/pjutil/pjutil.go
```diff
@@ -85,10 +81,10 @@ func createRefs(pr *scm.PullRequest, baseSHA string) plumber.Refs {
 	}
 }
 
-// NewPresubmit converts a config.Presubmit into a builder.PlumberJob.
+// NewPresubmit converts a config.Presubmit into a builder.PlumberArguments.
 // The builder.Refs are configured correctly per the pr, baseSHA.
 // The eventGUID becomes a github.EventGUID label.
-func NewPresubmit(pr *scm.PullRequest, baseSHA string, job config.Presubmit, eventGUID string) plumber.PlumberJob {
+func NewPresubmit(pr *scm.PullRequest, baseSHA string, job config.Presubmit, eventGUID string) plumber.PlumberArguments {
```
ccojocar: What is a `presubmit` in the context of `jenkins-x` pipeline.

pkg/plumber/constants.go
jstrachan: sounds good to me - PR on way shortly...

pkg/plumber/types.go
jstrachan: on way...

pkg/plumber/types.go
jstrachan: I figured at some point we may wanna add a concurrently limit - but lemme zap for now

pkg/prow/pjutil/pjutil.go
jstrachan: it a PR pipeline as opposed to a release pipeline - these names match the current prow config we use

pkg/prow/config/config.go
jstrachan: this PR doesn't touch the prow config - I figured we'd tackle cleaning that up in a separate PR

pkg/plumber/types.go
jstrachan: removed

jstrachan:
@ccojocar think I've fixed all those issues you found

pkg/plumber/types.go
```diff
@@ -7,76 +7,41 @@ import (
...
-// PlumberJobType specifies how the job is triggered.
-type PlumberJobType string
+// PipelineKind specifies how the job is triggered.
+type PipelineKind string
 
 // Various job types.
 const (
 	// PresubmitJob means it runs on unmerged PRs.
-	PresubmitJob PlumberJobType = "presubmit"
+	PresubmitJob PipelineKind = "presubmit"
```
ccojocar: Shouldn't we change `presubmit`, `postsubmit`, `periodic` and `batch` to jenkins-x pipeline specific values: `pullRequest`, `release`, `feature`. We will map the webhook events directly to these and not rely anymore on prow terminology as a middle layer.

jstrachan: I thought we decided for now keep the prow configuration format; so I figured leaving the `presubmit/postsubmit` terms until we revamp the config to minimise confusion.

I figure we can replace the prow config + the presubmit/postsubmit terms at a later date?

ccojocar: Agree. Let's keep the existing prow types and names to avoid confusion until we do a more radical improvement in the config.

jstrachan:
@ccojocar is there anything I've missed from your review?

Issue #94, opened 2019-08-05
webhook handler is firing 2x per event


wbrefvem:
I have vague recollections of this happening once before. Basically a single webhook from GitHub is getting processed twice. This causes the bdd tests to fail, since, e.g. in the golang http test, the expectation in at least one place is that only a single build exists.

wbrefvem:
This maybe PEBKAC and/or some weirdness with the install, but regardless I'll post findings here for visibility.

wbrefvem:
Looks like the image tag in the chart is a bit stale, `0.0.31` as opposed to `0.0.64`, the latest. Bumping the image fixes the issue.

PR #95, opened 2019-08-05
fix: bump lighthouse image tag in chart

wbrefvem:
Addresses #94

### DIFF
--- a/charts/lighthouse/values.yaml
+++ b/charts/lighthouse/values.yaml
@@ -12,7 +12,7 @@ hmacToken: ""
 replicaCount: 1
 image:
   repository: gcr.io/jenkinsxio/lighthouse
-  tag: 0.0.31
+  tag: 0.0.64
   pullPolicy: IfNotPresent
 
 # define environment variables here as a map of key: value


jstrachan:
the image tag is already updated as part of the release pipeline:
https://github.com/jenkins-x/lighthouse/blob/v0.0.65/charts/lighthouse/values.yaml#L15

wbrefvem:
@jstrachan Okay so this is something I didn't understand until just now. I get that polluting master with a new commit every time the version bumps is not ideal, but the way things now work we 1) leave master to languish with an old version, and 2) don't properly tag master with each release. Both make developing worse, the latter because local git tooling (not just the git cli, but shell integration as well) is a lot nicer when you can see all of the tags in master's `git log`.

jstrachan:
we usually use `latest` in master - how about using that in the master branch and getting the pipeline to also tag each release image with `latest`?

I don't follow the tag comment though - everything's tagged here just fine: https://github.com/jenkins-x/lighthouse/releases e.g. https://github.com/jenkins-x/lighthouse/tree/v0.0.65

PR #100, opened 2019-08-09
fix: switch the docker image to use alpine and cleanup the makefile

ccojocar:
By switching from centos to alpine has reduced the image size from 600MB to 100MB.

part of #77


jstrachan:
@ccojocar did the new dockerfile work? We need a working git install for chatops unless we move to a pure Golang git provider. It would be good to double check the new image with `/lgtm / /approve` etc

ccojocar:
>  did the new dockerfile work? We need a working git install for chatops unless we move to a pure Golang git provider. It would be good to double check the new image with /lgtm / /approve etc

There is a working `git` installed see L11 in the Dockerfile. I did some basic test locally but I would need still to do more validation. 

Yeah the plan is to use `go-git` in the mid-term and then switch to `scratch` as base. The images is still 100MB due to git and other related dependencies installed by alpine.

ccojocar:
@wbrefvem Please have a look over this PR? Thanks!

Issue #81, opened 2019-08-05
Implement a clean integration with jx meta-pipeline


ccojocar:
Integrate the jx meta-pipeline client from #61 in Ligthouse. 

Use meta-pipeline client into the Plumber instead of calling the jx command options. 

https://github.com/jenkins-x/lighthouse/blob/e3d11e12ff05f7d086a8e90a8050431b647ebe8e/pkg/plumber/plumber.go#L32

The meta-client interface can be found here `pkg/tekton/metapipeline/client.go` and see the PR which implemented the client https://github.com/jenkins-x/jx/pull/5051. An example is already available in `jx` in pipelinerunner `https://github.com/jenkins-x/jx/pull/5051/files#diff-cfadf679d93b353852ef9097113f3e53`

ccojocar:
@jstrachan I move the issue in progress and assigned it to you.

PR #104, opened 2019-08-24
WIP: fix: Switch to metapipeline client

abayer:
This is...well, an attempt. I'm not sure it's right, I'm not sure it'll fix the `jenkins-x-versions` `boot-lh` and `boot-lh-ghe` tests (which are failing at least since the metapipeline client was introduced), but it's something?

I can't for the life of me figure out how Lighthouse is updating the status of a running pipeline on a PR, though, which is the real problem in the `jenkins-x-versions` tests.

/assign @wbrefvem 
/assign @ccojocar 

fixes: #81

### DIFF
--- a/pkg/plumber/plumber.go
+++ b/pkg/plumber/plumber.go
@@ -7,8 +7,7 @@ import (
 
 	"github.com/jenkins-x/go-scm/scm"
 	"github.com/jenkins-x/jx/pkg/cmd/opts"
-	"github.com/jenkins-x/jx/pkg/cmd/step/create"
-	"github.com/jenkins-x/jx/pkg/prow"
+	"github.com/jenkins-x/jx/pkg/tekton/metapipeline"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
@@ -32,16 +31,16 @@ func NewPlumber(repository scm.Repository, commonOptions *opts.CommonOptions) (P
 func (b *PipelineBuilder) Create(request *PipelineOptions) (*PipelineOptions, error) {
 	spec := &request.Spec
 
-	pullRefData := b.getPullRefs(spec)
-	pullRefs := ""
-	if len(spec.Refs.Pulls) > 0 {
-		pullRefs = pullRefData.String()
-	}
-
 	repository := b.repository
 	name := repository.Name
 	owner := repository.Namespace
 	sourceURL := repository.Clone
+
+	pullRefData := b.getPullRefs(sourceURL, spec)
+	pullRefs := ""
+	if len(spec.Refs.Pulls) > 0 {
+		pullRefs = pullRefData.String()
+	}
 
 	branch := b.getBranch(spec)
 	if branch == "" {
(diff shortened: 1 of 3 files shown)


pkg/plumber/plumber.go
```diff
@@ -66,23 +71,36 @@ func (b *PipelineBuilder) Create(request *PipelineOptions) (*PipelineOptions, er
...
+		Context:      spec.Context,
+		// No equivalent to https://github.com/jenkins-x/jx/blob/bb59278c2707e0e99b3c24be926745c324824388/pkg/cmd/controller/pipeline/pipelinerunner_controller.go#L236
+		//   for getting environment variables from the prow job here, so far as I can tell (abayer)
+		// Also not finding an equivalent to labels from the PipelineRunRequest
+		ServiceAccount: sa,
+		// I believe we can use an empty string default image?
+		DefaultImage: "",
+	}
+
+	metaPipelineClient, err := metapipeline.NewMetaPipelineClient()
```
hferentschik: I don't know the whole context of how this is exactly called, but I assume this is called on each pipeline trigger. Ideally one wants to cache the meta pipeline client opposed to creating one on request. I am not familiar with the rest of the code, but is this Plumber also called on each request? Or is it kept around? If so, one could probably create the client as part of the plumber.

hferentschik: BTW, I still believe that architecturally we would be better off not depending directly on jx and the meta pipeline client. Instead we should be using the pipelinerunner and call it from Lighthouse.

ccojocar: I think this is a good point. it seems to be executed on every request. See the https://github.com/jenkins-x/lighthouse/blob/e3d11e12ff05f7d086a8e90a8050431b647ebe8e/pkg/prow/plugins/trigger/trigger.go#L231.

ccojocar:
> I can't for the life of me figure out how Lighthouse is updating the status of a running pipeline on a PR, though, which is the real problem in the jenkins-x-versions tests.

`crier` needs to be installed as well. Lighthouse stand alone does not have yet capability to update the status of a PR, though there is something when a a job is skipped by the trigger (See https://github.com/jenkins-x/lighthouse/blob/e3d11e12ff05f7d086a8e90a8050431b647ebe8e/pkg/prow/plugins/trigger/trigger.go#L251).

abayer:
fwiw, I'm 100% fine with @wbrefvem taking this and doing whatever's appropriate to it. I was just experimenting when I thought that we couldn't get the LH BDD tests on `jenkins-x-versions` to pass. =)

hferentschik:
> fwiw, I'm 100% fine with @wbrefvem taking this and doing whatever's appropriate to it. I was just experimenting when I thought that we couldn't get the LH BDD tests on jenkins-x-versions to pass. =)

Sure. No worries. Just saying the client is designed to be "cached". It probably does not matter much, but if you create the client on each request we need to make the calls to instantiate the (Kube) clients over and over again. That might not make matter much for our use case, but it's worth thinking about imo.

jstrachan:
@abayer is this PR still needed - looks like we are using the metapipeline client with the client passed in now? 
https://github.com/jenkins-x/lighthouse/blob/master/pkg/plumber/plumber.go#L31

abayer:
Yup, sorry, forgot to close!

PR #105, opened 2019-08-27
fix: Switch to metapipeline client

wbrefvem:
This includes support for using the plugin agent to access a cached metapipeline client, although this approach is not implemented in all cases.

### DIFF
--- a/pkg/plumber/plumber.go
+++ b/pkg/plumber/plumber.go
@@ -7,8 +7,7 @@ import (
 
 	"github.com/jenkins-x/go-scm/scm"
 	"github.com/jenkins-x/jx/pkg/cmd/opts"
-	"github.com/jenkins-x/jx/pkg/cmd/step/create"
-	"github.com/jenkins-x/jx/pkg/prow"
+	"github.com/jenkins-x/jx/pkg/tekton/metapipeline"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
 )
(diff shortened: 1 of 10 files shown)


pkg/plumber/plumber.go
```diff
@@ -66,23 +71,31 @@ func (b *PipelineBuilder) Create(request *PipelineOptions) (*PipelineOptions, er
...
+	pipelineCreateParam := metapipeline.PipelineCreateParam{
+		PullRef:      pullRefData,
+		PipelineKind: kind,
+		Context:      spec.Context,
+		// No equivalent to https://github.com/jenkins-x/jx/blob/bb59278c2707e0e99b3c24be926745c324824388/pkg/cmd/controller/pipeline/pipelinerunner_controller.go#L236
+		//   for getting environment variables from the prow job here, so far as I can tell (abayer)
+		// Also not finding an equivalent to labels from the PipelineRunRequest
+		ServiceAccount: sa,
+		// I believe we can use an empty string default image?
+		DefaultImage: "",
```
hferentschik: Jupp, or even better don't set it at all

pkg/plumber/plumber.go
```diff
@@ -66,23 +71,31 @@ func (b *PipelineBuilder) Create(request *PipelineOptions) (*PipelineOptions, er
...
 	}
-	po.CommonOptions = b.commonOptions
-	po.ServiceAccount = sa
 
-	err := po.Run()
+	pipelineCreateParam := metapipeline.PipelineCreateParam{
+		PullRef:      pullRefData,
+		PipelineKind: kind,
+		Context:      spec.Context,
+		// No equivalent to https://github.com/jenkins-x/jx/blob/bb59278c2707e0e99b3c24be926745c324824388/pkg/cmd/controller/pipeline/pipelinerunner_controller.go#L236
```
hferentschik: Right, so I would think that there are a few pieces (labels, env variables) missing here to make sure that jobs triggered by Lighthouse behave the same as jobs triggered by Prow

hferentschik:
> @abayer Looks good to me

In terms of the meta pipeline client and it not being recreated all the time this looks ok to me. As @abayer pointed out there seems to be some pieces missing to really make this work though.

PR #354, opened 2019-11-21
feat(tests): Add BDD tests for GHE, separate prow contexts

### LINKED ISSUE #355: Define contexts for BDD tests
In order to prevent bloat in `jenkins-x.yaml`, we should define separate prow contexts for running BDD tests against GitHub, GHE, etc.

wbrefvem:
This is meant to test the basic configuration of the GHE BDD tests, but eventually they will have a separate prow contexts.

Fixes #355


abayer:
Mentioned on slack, but just to be safe - gotta name the new context files `jenkins-x-ghe.yml` etc - the file extension does explicitly have to be `.yml`, not `.yaml`.

abayer:
@wbrefvem you need to be building the images in the new contexts - I.e., you need https://github.com/jenkins-x/lighthouse/blob/687152bc2949c3a6978f4acb6d145fda8c693a1c/jenkins-x.yml#L83-L113 in their pipelines.

abayer:
(That’s why the new contexts time out waiting for the Lighthouse pod to be running - there’s no image for the version specified.)

abayer:
/lgtm

I officially declare this done. =)

PR #470, opened 2020-01-17
fix: Use token secret like Prow does

### LINKED ISSUE #465: lighthouse chart uses jx-pipeline-git-(kind)-(name) secret in deployments but secret isn't available
Since the `jx-auth-config` configmap started getting created, a boot+gitops cluster won't create secrets like `jx-pipeline-git-(kind)-(name)`...which is a problem for Lighthouse, which uses the following in its deployment configuration for its pods:

```
        - name: "GIT_USER"
          valueFrom:
            secretKeyRef:
              name: "jx-pipeline-git-{{ .Values.git.kind }}-{{ .Values.git.name }}"
              key: username
        - name: "GIT_TOKEN"
          valueFrom:
            secretKeyRef:
              name: "jx-pipeline-git-{{ .Values.git.kind }}-{{ .Values.git.name }}"
              key: password
          value: "{{ .Values.git.token }}"
```

@ccojocar - what should we be doing instead in this scenario?

abayer:
Down the road, this should change to pulling the token from vault dynamically, but we need the vault mutating webhook to make that work, so for now...tada.

Also, turns on vault and gitops for all contexts.

fixes #465

### DIFF
--- a/bdd/bbs/ci.sh
+++ b/bdd/bbs/ci.sh
@@ -43,10 +43,21 @@ export JX_DISABLE_TEST_CHATOPS_COMMANDS="true"
 # TODO temporary hack until the batch mode in jx is fixed...
 export JX_BATCH_MODE="true"
 
+# Push the snapshot chart
+pushd charts/lighthouse
+make snapshot
+popd
+
+# Use the latest boot config promoted in the version stream instead of master to avoid conflicts during boot, because
+# boot fetches always the latest version available in the version stream.
+git clone https://github.com/jenkins-x/jenkins-x-versions.git versions
+export BOOT_CONFIG_VERSION=$(jx step get dependency-version --host=github.com --owner=jenkins-x --repo=jenkins-x-boot-config --dir versions | sed 's/.*: \(.*\)/\1/')
 git clone https://github.com/jenkins-x/jenkins-x-boot-config.git boot-source
-cp bdd/bbs/jx-requirements.yml boot-source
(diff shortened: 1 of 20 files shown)


abayer:
Latest problem is https://github.com/jenkins-x/jx/issues/6568

abayer:
/assign @jstrachan 
/assign @mrageh 

Once https://github.com/jenkins-x/jenkins-x-versions/pull/779 is merged, I think this is good to go, with boot/gitops/vault/chatops all actually working right for GHE and github.com, and boot/gitops/vault working for Gitlab. The chatops test doesn't work right with either gitlab or bitbucket at the moment, but I'll open issues for those.

abayer:
Gitlab's gonna fail - it didn't set a status on the second staging PR. I may just revert the gitlab test for now.

abayer:
Created https://github.com/jenkins-x/lighthouse/issues/477 for the Bitbucket failure.

abayer:
Removed my hack of using the newer `jenkins-x-platform` now that it's in the version stream - github was gonna hang on not updating the result of a staging PR, which is similar to what gitlab was hanging on (though there it was earlier and not even creating the pending status). I think that may be flaky in general, so let's see what happens this time on all three, since in both github and gitlab, _most_ PRs had statuses set correctly. I'm willing to live with it being a little flaky right now as long as it's not consistent in any of the three providers and we open a followup issue.

PR #481, opened 2020-01-21
fix: Rerun PipelineRuns failed due to known race conditions

### LINKED ISSUE #480: Pipeline not found (and Task not found) errors not resulting in new runs being kicked off
This is an inherent flaw in LH as it stands right now vs Prow - because there's nothing monitoring pipelines like the ProwJob controller in Prow, there's nothing to tell when a `PipelineRun` hits the "not found" race conditions and then rerunning them. This is...not ideal. Not yet sure what the answer is.

abayer:
Similar to fixes made to the Jenkins X fork of Prow, but leveraging
the existing Tide pod to check for such runs and rerun them. We add a
label to each failed `PipelineRun` that we rerun so that they won't
get rerun multiple times.

fixes #480


abayer:
There’s still a problem with PR builds sometimes not getting marked as completed by the build controller, and therefore never reporting status, but that’s a separate issue.

PR #534, opened 2020-02-24
fix: retest and other checks need pr.Head.Ref, not pr.Base.Ref

abayer:
This is resulting in retest looking at contexts on master, not on the
PR, which is wrong and results in everything getting rebuilt.

fixes https://github.com/jenkins-x/jx/issues/6796


abayer:
I’m going to follow this up tomorrow with a BDD test to verify it going forward.

Issue #545, opened 2020-02-27
Support the the same chatops behavior on issues that Prow does


abayer:
I need to dig into this a bit more, but Prow does support some of its commands on either a PR or an issue, like:

* `/retitle`
* `/assign`, `/unassign`
* `/meow`, `/woof`, and the like
* `/help`, `/good-first-issue` (and their `remove-` variants)
* The label helpers (`/kind`, `/area`, etc, and their `remove-` variants)
* `/close`, `/reopen`
* `/lifecycle`
* `/milestone`
* `/project`

PR #546, opened 2020-02-27
chore: Decrease verbosity of logging for unknown webhook kinds

abayer:
The current verbosity overwhelms the logs thanks to issue and status
hooks, which we don't currently handle. Since we just use the exact
same hook configuration on GitHub as Prow would, we're getting those
whether we want them or not, and with #545 eventually we'll want
them. But for now, let's only vomit out all that noise on debug.


abayer:
can I comment?

PR #560, opened 2020-02-28
fix: Override shouldn't create new builds

### LINKED ISSUE #557: /override creates a perpetual loop of new runs of the overridden context
Not yet sure what exactly is going on, but yeah, override doesn't work. The context is briefly marked as overridden (no comment, though, just the status update), and then it kicks off a new run, then it overrides that run, then it kicks off another run, etc forever.

abayer:
This was an artifact of Prow overriding by creating a dummy job that
instantly passed, while we handle that directly.

fixes #557


abayer:
Verified this one - it works right.

PR #578, opened 2020-03-05
fix: overridden status takes precedence over latest build result

### LINKED ISSUE #557: /override creates a perpetual loop of new runs of the overridden context

abayer:
This gets the user-facing behavior to more closely match Prow - `/override` creates a commit status with an `Overridden by ...` description, and even if the latest `PipelineActivity` for that context is failed or pending and hasn't reported since the `/override` went in, Tide should treat this context as successful/ready to merge.

In Prow, `/override` creates a new `ProwJob` - one that doesn't actually kick off a pipeline but just sets the commit status and sets its own status to successful. Since Prow uses `ProwJob`s as the "build status" source, where Lighthouse currently uses `PipelineActivity`s, Prow's Tide will see the override-created and successful `ProwJob` as the latest for the given context, rather than the created-earlier failed or pending `ProwJob`, and the PR (assuming all contexts are successful, of course!) will happily merge.

In Lighthouse, we don't create a `PipelineActivity` via `/override`, because `PipelineActivity` doesn't just serve Lighthouse's purposes - its main purpose is build status/reporting/etc for Jenkins X, so we really don't want to mess around with it. Without this change, we'd just see the latest failed or pending `PipelineActivity` for the PR+context, and then things would get silly and it would insist on running a new build. With this change, we check the commit status for the context to see if it's successful and starts with `Overridden by`, and if so, we override the `PipelineActivity`-driven status with success.

This will all probably end up getting wiped away in the future when we rearchitect to have a `ProwJob` equivalent (which I think we also need for batch - it's possible this change _may_ fix batch, but I don't think it will), but for the time being, I'd really like `/override` to work. =)

fixes #557

### DIFF
--- a/bdd/github/ci.sh
+++ b/bdd/github/ci.sh
@@ -85,7 +85,7 @@ jx step bdd \
     --no-delete-app \
     --no-delete-repo \
     --tests test-create-spring \
-    --tests test-quickstart-golang-http \
+    --tests test-lighthouse \
     --tests test-app-lifecycle
 
 bdd_result=$?
--- a/bdd/github/jx-requirements.yml
+++ b/bdd/github/jx-requirements.yml
@@ -1,5 +1,7 @@
 cluster:
   clusterName: lh-bdd-github
+  devEnvApprovers:
+    - jenkins-x-bot-test
   environmentGitOwner: jenkins-x-versions-bot-test
   project: jenkins-x-bdd3
   provider: gke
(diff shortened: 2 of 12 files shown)


abayer:
This is now ready to go!

PR #590, opened 2020-03-10
fix: Add support for GitHub App authentication to webhook controller

### LINKED ISSUE #586: Lighthouse webhook controller should support GitHub App-generated tokens
We already do in the Tide controller, so let's do so in the webhook controller as well.

abayer:
This takes how Tide was already doing this and makes it reusable, with some tweaks for the webhook controller needing to already have an instantiated scm client for webhook parsing, before it knows the owner, and makes sure we're following the same code path for scm provider client creation and authorization everywhere.

Also, this reworks the charts a bit to include the relevant secret and configuration in the webhook controller deployment as well.

fixes #586


abayer:
This is very much WIP - I need to figure out a good way to actually test the GitHub App behavior in the BDD tests, but I think this is generally on the right path.

PR #603, opened 2020-03-19
feat: Move status reporting into new controller

### LINKED ISSUE #437: Move status updates from jx build controller to Lighthouse
This is a bit of a thorny issue - there are some concerns I can't go into detail on here. But the goal here is to stop doing the status updates at the end of a pipeline's execution in jx's build controller and instead do it in LH. This would be much cleaner, and I think would be better overall as a result of colocating all the GitHub/BitBucket/Gitlab/etc API interactions here.

I'm not sure yet what the implementation would entail - if we want to go in the direction of making Lighthouse able to handle triggering things besides just jx pipelines, we'd definitely need a `ProwJob` equivalent here, but my instinct is to lean into the jx-specificness of LH and instead add a field to `PipelineActivity` named something like `Reported`, and then watch for `PipelineActivity`s in LH that have a non-nil `CompletedTimestamp` and `Reported` as false, reporting status for such `PipelineActivity`s and updating their `Reported` field.

abayer:
Rather than using the jx build controller for status reporting, this uses a new CRD, `LighthouseJob`, and a new controller, `lighthouse-foghorn`, for recording what has been kicked off, status, whether it's been reported, etc. This is more in line with how Prow works, and allows things like override to work without hacks.

We also rename tide to `lighthouse-keeper` here, because hey, I'm on a role of renaming things.

fixes #437


abayer:
This replaces various earlier WIP PRs with a new WIP PR! Since #590 has merged, it made sense to squash this down and rebase it properly.

A fair amount more work is needed - unit tests for the foghorn controller, some more cleanup of dead/poorly-named code, and I want to run this for a while on a testbed cluster to make sure it's not too chatty/eager to act on `PipelineActivity`s. Also I need to put in GC logic for no-longer-relevant `LighthouseJob`s - I'm leaning towards deleting them like 7 days after they're either no longer the most recent build for the PR/context or their PR has merged, and 30 days otherwise. To be determined, but definitely a requirement for merging this.

abayer:
GC cronjob in - it's currently very stupid, just nuking everything 7+ days old, but we can circle back on that.

abayer:
Tests may fail because I removed a bit I had in the BDD tests to force disabling status reporting from the build controller, because I want to be 100% sure the "disable if there's a `lighthouse-foghorn` deployment" logic I added to the build controller works correctly. And that _just_ made it into the version stream, and I may have pushed here a little prematurely. =) If so, I'll just rerun them and they should be fine.

PR #619, opened 2020-03-28
chore: allow log format to be configured

garethjevans:


abayer:
So this is fine, but I don't think we actually use that field in any meaningful way. Might as well merge and follow up later with a general rewrite of logging.

PR #660, opened 2020-04-21
fix: Update status immediately after creating pipeline

abayer:
The delay between `/retest` and the previous bad status being replaced is annoyingly long, so let's do that quicker.

### DIFF
--- a/pkg/plugins/trigger/generic-comment_test.go
+++ b/pkg/plugins/trigger/generic-comment_test.go
@@ -110,13 +110,14 @@ func TestHandleGenericComment(t *testing.T) {
 			ShouldBuild: false,
 		},
 		{
-			name:        "accept /test from non-trusted member if PR author is trusted",
-			Author:      "untrusted-member",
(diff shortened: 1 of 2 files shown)


abayer:
The gitlab failure is real - for some reason, this results in the status never getting updated past pending for gitlab.

pkg/plugins/trigger/generic-comment_test.go
```diff
@@ -776,119 +788,121 @@ func TestHandleGenericComment(t *testing.T) {
...
-					Number: 0,
-					Head: scm.PullRequestBranch{
-						Sha: "cafe",
-					},
-					Base: scm.PullRequestBranch{
-						Ref: tc.Branch,
-						Repo: scm.Repository{
-							Namespace: "org",
-							Name:      "repo",
+		t.Run(tc.name, func(t *testing.T) {
```
mrageh, line 791: Is `t.Run` a function that takes the name and an anonymous function?

PR #661, opened 2020-04-22
feat: make team name configuable in the report url

garethjevans:


pkg/foghorn/controller.go
```diff
@@ -422,9 +429,14 @@ func (c *Controller) getReportURLBase() string {
 	return os.Getenv("LIGHTHOUSE_REPORT_URL_BASE")
 }
 
+// getReportURLTeam gets the team to construct the report url
+func (c *Controller) getReportURLTeam() string {
+	return os.Getenv("LIGHTHOUSE_REPORT_URL_TEAM")
```
mrageh, line 434: What happens if this env variable is not set?

Should we guard against that or will it always be present?

pkg/foghorn/controller.go
```diff
@@ -384,6 +384,13 @@ func (c *Controller) reportStatus(ns string, activity *jxv1.PipelineActivity, jo
 	}
 	urlBase := c.getReportURLBase()
 	if urlBase != "" {
+		urlTeam := c.getReportURLTeam()
+		team := ns
+		// override with env var if set
```
mrageh, line 389: Why do we override with env var if it set?

abayer: Because the namespace is the proper default.

PR #676, opened 2020-04-22
fix: Check org members, not just collabs, for approve rights

### LINKED ISSUE #675: OWNERS not being read properly for some users
I'm really not sure what's causing this yet, but Gareth and I can approve things in a number of repos but no one else can. I think this is a concurrency issue but am not sure yet.

abayer:
fixes #675

### DIFF
--- a/pkg/scmprovider/client.go
+++ b/pkg/scmprovider/client.go
@@ -47,6 +47,7 @@ type SCMClient interface {
 	// Functions implemented in organizations.go
 	ListTeams(string) ([]*scm.Team, error)
 	ListTeamMembers(int, string) ([]*scm.TeamMember, error)
+	ListOrgMembers(string) ([]*scm.TeamMember, error)
 
 	// Functions implemented in pull_requests.go
 	GetPullRequest(string, string, int) (*scm.PullRequest, error)
--- a/pkg/webhook/webhook.go
+++ b/pkg/webhook/webhook.go
@@ -2,7 +2,6 @@ package webhook
 
 import (
 	"fmt"
-	"io/ioutil"
 	"net/http"
 	"os"
 	"os/signal"
@@ -61,7 +60,7 @@ type Options struct {
 	botName          string
 	gitServerURL     string
 	configMapWatcher *watcher.ConfigMapWatcher
-	localGitDir      string
+	gitClient        git.Client
 }
 
 // NewCmdWebhook creates the command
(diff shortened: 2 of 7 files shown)


pkg/webhook/webhook.go
```diff
@@ -120,10 +119,18 @@ func (o *Options) Run() error {
...
 	}
 
-	o.localGitDir, err = ioutil.TempDir("", "git")
+	gitClient, err := git.NewClient(o.gitServerURL, o.gitKind())
 	if err != nil {
-		return errors.Wrapf(err, "failed to create git temp directory")
+		logrus.WithError(err).Fatal("Error getting git client.")
 	}
+	defer func() {
+		err := gitClient.Clean()
```
hferentschik, line 127: I guess I am a bit surprised to see a `Clean` here. What does a git client have to clean up?
Also, this is a controller, right and we are in the `Run` method, so this defer will execute when the whole controller dies. So is this just a cleanup to do things "right" or does this really try to fix anything and if so what?

hferentschik: Lighthouse is really too much of a black box for me atm. I will get a checkout and start poking around a bit. Then I will be able to give better feedback.

pkg/scmprovider/client.go
```diff
@@ -47,6 +47,7 @@ type SCMClient interface {
 	// Functions implemented in organizations.go
 	ListTeams(string) ([]*scm.Team, error)
 	ListTeamMembers(int, string) ([]*scm.TeamMember, error)
+	ListOrgMembers(string) ([]*scm.TeamMember, error)
```
hferentschik, line 50: Not related, but the one function which sticks our is `DeleteStateComments`. Where is this coming from? I would have loved to see `StaleComments` and `DeleteComments`. But that's besides the point for this PR.

pkg/webhook/webhook.go
abayer: So fwiw, the "git client" here needs to clean up because it mainly just does clones of repos to check their `OWNERS` files, so in theory when the process dies it should nuke the temp dir. The problem that this change is fixing is that we were doing that wayyyyyyy too much, in fact. So yeah, here, now that I've made it more like Prow behavior (just initing the git client once, not initing it each time we process a webhook), the cleanup should just be "doing things right".

pkg/scmprovider/client.go
abayer: https://github.com/jenkins-x/lighthouse/blob/master/pkg/scmprovider/issues.go#L126 fwiw

PR #689, opened 2020-04-24
feat: GitLab support in Keeper

### LINKED ISSUE #688: Support for non-GraphQL providers in Keeper
Keeper is built around GitHub's GraphQL API, which isn't the same as GitLab's (very limited) GraphQL API, let alone other providers that may not have GraphQL APIs in the first place. So we should have a fallback for other providers that will allow Keeper to actually work.

Gonna start with GitLab.

abayer:
With this change, any SCM provider that implements the needed REST queries (and populates `scm.Repository` and `scm.PullRequest` properly) should work. The initial target for this is GitLab.

Due to GitLab quick actions hijacking `/approve` and `/assign` most notably (there are some other quick action commands that overlap), I've added the ability to invoke any Lighthouse command with `/lh-(whatever)` as well as `/whatever`.

fixes #688

### DIFF
--- a/bdd/gitlab/ci.sh
+++ b/bdd/gitlab/ci.sh
@@ -70,8 +70,7 @@ echo "Building lighthouse with version $VERSION"
 helm init --client-only
 helm repo add jenkins-x https://storage.googleapis.com/chartmuseum.jenkins-x.io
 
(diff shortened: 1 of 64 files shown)


abayer:
/hold

The now fixed `/cc` stuff isn't working for gitlab.

PR #704, opened 2020-05-01
fix: Get all pages of results for provider queries

### LINKED ISSUE #703: Provider queries don't get all pages
We should be getting all pages for lists.

abayer:
fixes #703


abayer:
So I'm not _happy_ about this approach - I'd like to actually build handling of pagination into go-scm directly, but I failed to find an "easy" way to do that. So this will work for the moment.

PR #763, opened 2020-05-26
chore: remove use of jx step credential

cagiti:


abayer:
Hrm, something's borked in the `jenkins-x.yml` pipeline as a result of this. I'll try to take a look in a bit.

abayer:
Fixed it - I removed a bunch of stuff we didn't need, but the problem was having two secrets both using `mountPath: /secret`. Oops.

cagiti:
> using `mountPath: /secret`. Oops

Ah cheers Andrew, was coming back to this tonight... going through my open PRs 👍

cagiti:
@abayer : was hoping we could remove this also https://github.com/jenkins-x/lighthouse/blob/1e5c03139760628f745f32bed0381344f1f9d00e/jenkins-x.yml#L134-L145 but from a brief investigation I can't see how it's being used so wondered whether `/builder/home/git-token` was interpreted as `~/git-token` and a default location for the git token file?

PR #774, opened 2020-05-29
chore: add extra logging when determining if builds should trigger

garethjevans:


abayer:
The GitLab failure is weird - I'm running a couple other PRs against GitLab to verify, but it appears to be flakiness on their side when commenting on a merge request, since it's failed with a 500 on two attempts to make different comments, sometimes succeeding and sometimes failing.

abayer:
Of the three GitLab runs on three different PRs, one passed and two failed, so I’m fairly certain this is just flaky on their side right now.

PR #778, opened 2020-06-03
fix: missing ServiceAccount for gc-jobs

vbehar:


abayer:
/lgtm

Whoops. =) Good catch!

PR #780, opened 2020-06-03
fix: Correct path for gc-jobs command in cronjob

abayer:
Stumbled across this after I reviewed #778.

fyi @vbehar. =)


vbehar:
thanks @abayer, it's working now - well, not doing anything for the moment because we have a maxAge of 720h, but at least the pod is running

PR #842, opened 2020-07-06
WIP: chore: Decouple jx from Lighthouse

abayer:
We're still going to be building something that depends on jx, but the end result will be a separate controller that handles triggering metapipelines (in response to a `LighthouseJob`) and updating `LighthouseJob`s (from listening to `PipelineActivities`).


abayer:
Gitlab isn't getting the labeling done properly - some weird side effect, but I haven't nailed down what it is yet.

abayer:
Turns out the gitlab problem is one of the two users we use in it has been locked somehow. So the test is gonna keep failing until it gets fixed. D'oh!

cmd/jxcontroller/main.go
```diff
@@ -0,0 +1,112 @@
...
+
+	jxclient "github.com/jenkins-x/jx-api/pkg/client/clientset/versioned"
+	jxinformers "github.com/jenkins-x/jx-api/pkg/client/informers/externalversions"
+	clientset "github.com/jenkins-x/lighthouse/pkg/client/clientset/versioned"
+	lhinformers "github.com/jenkins-x/lighthouse/pkg/client/informers/externalversions"
+	"github.com/jenkins-x/lighthouse/pkg/clients"
+	"github.com/jenkins-x/lighthouse/pkg/interrupts"
+	"github.com/jenkins-x/lighthouse/pkg/jx"
+	"github.com/jenkins-x/lighthouse/pkg/logrusutil"
+	"github.com/sirupsen/logrus"
```
cagiti, line 18: maybe use github.com/jenkins-x/jx-logging/pkg/log here?

abayer: Good call - will do.

cagiti:
@abayer: overall seems really good to me, I know you've highlighted missing tests etc... but a good first step :)
Seems it might help towards moving the metaclient into its own repository? Let me know after you've added a final update and i'll take a quick look 👍

abayer:
@cagiti Regrettably, this doesn't make much of a difference on the jx side in terms of decoupling the metapipeline stuff - that's still a messy challenge that I poked at a bit before going this direction instead.

abayer:
Closing for a new cleaner PR.

PR #845, opened 2020-07-10
Fixing some 404s

jglick:
Looks like you got rid of the `prow/` prefix on packages at some point?

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ The following environment variables are used:
 
 ## Features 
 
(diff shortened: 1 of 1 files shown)


abayer:
@jglick Could you please update the commit message to be "fix: ..."? Thanks.

jglick:
Or `chore:`? /me sighs

abayer:
Yeah, either one works.

PR #850, opened 2020-07-16
feat: Decouple jx interaction into a separate controller

abayer:
This will allow us to eventually split out the jx controller into a separate repository and chart, and more importantly, it'll let us add additional controllers for other agents - most notably Tekton Pipeline and Jenkins.

This won't require any changes in jx, since the new controller is deployed and used by default.


jstrachan:
/lgtm

great work @abayer - I love being able to use lighthouse using other kinds of controller, like vanilla tekton

PR #855, opened 2020-07-20
WIP: feat: Support for triggering and reporting Tekton pipelines

abayer:
Just getting started - I'll end up throwing this PR away once the work is done

works toward #857


abayer:
First commit moves some stuff around, adds a few more labels to go onto `LighthouseJob`s (and eventually `PipelineRun`s), and adds the function to convert from a `PipelineRun` to an `ActivityRecord` with some tests. Next up is writing the controller and its tests, followed by writing an end-to-end test framework, which will do something like the following:

* Create a GKE cluster using `gcloud`
* Install Tekton (probably via the jx chart for the moment)
* Install the Lighthouse chart we build for the PR, using creds defined as secrets and an autogenerated HMAC token.
* Create a repository containing `OWNERS` and a bash script - the pipeline will end up just running that bash script, so when we want it to fail, that'll be easy enough to do.
* Create the `Pipeline`, `Task`, etc.
* Create the config for the repo in the `config` and `plugins` configmaps for Lighthouse, requiring lgtm and approve, with one context configured for presubmits using a `PipelineRunSpec` that just calls the bash script above.
* Add hooks for the repo pointing to our hooks ingress.
* From there, basically follow the same flow that we do in [the JX BDD Lighthouse tests](https://github.com/jenkins-x/bdd-jx/blob/master/test/suite/lighthouse/helpers.go) - i.e., open PR that'll fail, make sure the provider is updated to record the failure, make sure lgtm'ing as the "bot" user fails, approve as a second user (who'll be in `OWNERS` and added to the repo as a contributor when we create it), `/test this`, `/retest`, `/override` as the approver, wait for it to merge.

abayer:
So! I'm currently hijacking the `github` context for e2e tests against Tekton. I've got the tests working (latest commit adds the remainder, so there could be something there that doesn't work, but I think we should be good), so woooo.

Next up is figuring out how to deal with the git `PipelineResource` not being relevant any more - right now, I'm assuming the `Pipeline` has a `git-clone` task with certain parameter names on the `Pipeline` getting passed to `git-clone`, but I need to rework that to:
* Inspect the `Pipeline` and replace `git-clone` with `git-batch-merge` if we're building a PR, changing parameters as needed, and then switch from using a `PipelineRef` on the `PipelineRun` to an inline `PipelineSpec` with our modified `Pipeline`.
* Inspect the `Pipeline` and determine what the `Pipeline` parameters that get passed to `git-clone` are and change the parameters on the `PipelineRun` to match.
* Possibly add a `git-clone` task at the beginning of the `Pipeline` if there isn't one present, again using an inline `PipelineSpec` on the `PipelineRun`.

jstrachan:
> So! I'm currently hijacking the `github` context for e2e tests against Tekton. I've got the tests working (latest commit adds the remainder, so there could be something there that doesn't work, but I think we should be good), so woooo.

great!

> Next up is figuring out how to deal with the git `PipelineResource` not being relevant any more - right now, I'm assuming the `Pipeline` has a `git-clone` task with certain parameter names on the `Pipeline` getting passed to `git-clone`, but I need to rework that to:
> 
> * Inspect the `Pipeline` and replace `git-clone` with `git-batch-merge` if we're building a PR, changing parameters as needed, and then switch from using a `PipelineRef` on the `PipelineRun` to an inline `PipelineSpec` with our modified `Pipeline`.

wonder if it’s cleaner to have a single git clone task that can handle PR or release commands based on the parameters? Am just thinking that would simplify things from the lighthouse & tekton CRD perspective; less magic etc?

Another option could be to use conditionals to invoke one of two steps baked into pipelines based on the parameters. But given how common git stuff is; making a reusable step that did the right git thing for PRs or releases sounds like a nice way forward maybe?

> * Inspect the `Pipeline` and determine what the `Pipeline` parameters that get passed to `git-clone` are and change the parameters on the `PipelineRun` to match.
> * Possibly add a `git-clone` task at the beginning of the `Pipeline` if there isn't one present, again using an inline `PipelineSpec` on the `PipelineRun`.

jstrachan:
this is all great stuff. A tangential UX thought on this that in no way should delay/block this work...

editing PipelineRun resources with inline Pipeline / Tasks is a non trivial bit of YAML - we’d want to give developers the maximum amount of IDE / tooling / validation support working with those.

So I wonder if the default / recommended way to configure a PipelineRun for a presubmit / postsubmit / periodic in lighthouse should be via a URL or file name rather than embed the PipelineRun inside an already very complex blob of YAML (the LH/prow configuration).

Eg folks could store the PipelineRun as a YAML file inside their git repository somewhere and reference it via a file path; or could store it inside a separate git repository and reference it by URL.
Most IDEs / editors can do smart completion or nice validation of k8s CRDs with known schemas now; plus am sure we could add nice tekton CRD linters to repos to catch typos in changes to PipelineRun YAMLs ahead of time?

BTW I recently raised a PR to try do a similar thing for the LH configuration itself - so a simpler more CRD like configuration file could be stored inside a git repository for defining the preSubmits / postsubmits / periodic etc - which also helps simplify the configuration a little 
https://github.com/jenkins-x/lighthouse-config/pull/9

abayer:
@jstrachan - re `git-clone` etc - here I'm talking about [the `git-clone` task from the Tekton Catalog](https://github.com/tektoncd/catalog/tree/master/task/git-clone/0.1) (and its cousin [the `git-batch-merge` task from the catalog](https://github.com/tektoncd/catalog/tree/master/task/git-batch-merge/0.1). In the mid-term, getting a single catalog task that can handle both properly is probably the way to go, but I'm trying to get this to MVP at the moment, so I'm focusing solely on what I can do without needing changes upstream.

re: that lighthouse-config PR and config-in-repo - interesting ideas. Once the Tekton work is at MVP level and merged, I'd like to tackle a few other things first - separating out the jx controller into its own repo (and a new chart, `lighthouse-jx` probably, that we'll switch jx to use instead of `lighthouse`, with the new chart containing lighthouse and the jx controller, while the vanilla `lighthouse` chart doesn't have the jx controller - this is largely meant to get rid of the `github.com/jenkins-x/jx/v2` dependencies in `lighthouse.git` so that we can then look to move `lighthouse-config` back in here and generally tidy things up), and rethinking how we manage config in general, since the configmap approach is less ideal for non-jx use cases where the config/plugins aren't generated and managed automatically by jx. In-repo config would probably make sense to dig into after those things are done.

Also, I'm closing this PR today and replacing it with a clean PR that should be ready to merge.

PR #862, opened 2020-07-23
fix: message re approvers files

MarckK:
Signed-off-by: Kara de la Marck <MarckK@users.noreply.github.com>


MarckK:
This change is to fix #863.

PR #866, opened 2020-07-27
feat: Support for triggering and reporting Tekton pipelines

### LINKED ISSUE #857: Support Tekton Pipelines as execution engine
We support Jenkins X as an execution engine, so we should add other engines too! First up is Tekton Pipelines.

abayer:
Includes e2e test (with proper cleanup etc), autodetection of `git-clone` task in `Pipeline` and setting of `PipelineRun` parameters to match.

Caveats: we aren't yet handling having PR builds building against the PR branch merged to master, which still needs work, and build log capturing is still needed as well.

fixes #857


abayer:
This replaces #855

PR #868, opened 2020-07-27
chore: Remove lighthouse-jx-controller from repo

### LINKED ISSUE #867: Move jx controller to separate repo and chart
So this is going to be a multi-step process. We should move `pkg/jxcontroller`, `cmd/jxcontroller`, and the relevant stuff in `charts/lighthouse` into its own repository, with its own chart, add that chart to `jenkins-x-versions`, and then change `jenkins-x-boot-config` to depend on both `lighthouse` and `lighthouse-jx` (or whatever we name it) charts. Then jx.git can switch its PR creation on release to update the `lighthouse-jx` repo, and we can get rid of https://github.com/jenkins-x/lighthouse-config, since the reason for that in the first place was to get rid of the circular dependency scenario (where jx depends on the Lighthouse config stuff (replacing the Prow config dependency), but Lighthouse depends on jx (for what's now in the jx controller).

abayer:
Now that https://github.com/jenkins-x/lighthouse-jx-controller is a thing, once https://github.com/jenkins-x/jenkins-x-boot-config/pull/172 has merged and gotten into the version stream, we can remove `lighthouse-jx-controller` (and the `jx` dependency!) from Lighthouse completely.

fixes #867


abayer:
This is gonna fail initially. Which is fine! Once https://github.com/jenkins-x/jenkins-x-boot-config/pull/172 merges/gets in the version stream, this'll pass.

PR #897, opened 2020-08-10
fix: add a new repoconfig package and file format

jstrachan:
so that we can implement different strategies for configuring lighthouse such as:

* a git repository with global configuration using traditional prow/lh config and then a directory tree of repository specific configurations
* detect lighthouse configuration inside the actual git repositories themselves; so git repositories can configure their own presubmits/postsubmits like the GitHub Actions model

here is an example [lighthouse.yaml](https://github.com/jenkins-x/lighthouse/blob/repo-config/pkg/repoconfig/merge/test_data/simple/lighthouse.yaml) file to get the idea. The current global configuration for lighthouse is pretty big, complex and easy to get wrong - so the idea of this repository specific configuration is to make a single, simpler YAML format folks can use if they are not too worried about sticking with the classic global prow configuration format.

This new config file format aims to allow 1 repository to be configured in a smaller simpler configuration file using versioning + CRD-like yaml so it feels more natural to kubernetes developers.

Note that this PR doesn't actually change how lighthouse loads its configuration yet; its more just defining a repositories configuration file and helper methods for merging the repository-specific configuration file(s) with the global configuration.

In future PRs we can try introduce a strategy/interface for getting the config/plugins configuration for a repository and integrating it into the current lighthouse code so we can maybe have a couple of strategies for loading configuration (e.g. loading from a single git repository, loading from ConfigMaps or a global configuration + repository specific configurations inside git etc)


abayer:
I know this originally was written against the now defunct separate `lighthouse-config` repo - is that the only reason for `/pkg/repoconfig/v1alpha1/*_types.go`? Because my gut reaction would be to put them in `pkg/apis/lighthouse/v1alpha1`. Also, I'm thinking that it probably makes sense to solve the problem of extensibility of job config (so that, i.e., some other build engine could write a controller and contribute its own config fields in `JobBase` or `Presubmit` etc without having to have everything harcdoded in Lighthouse proper)...

jstrachan:
the main reason for the `v1alpha1` package versus using `config` was so that we can version this configuration file going forward; there's no real easy way to retrofit versioning into the classic prow configs.

I did ponder putting them in `pkg/apis/lighthouse/v1alpha1` - I originally did that in lighthouse-config - but I wondered if folks would get confused if they were mostly used as files in git rather than CRDs; but since they are designed to look + feel CRD like thats maybe not a biggie. Plus folks could use them as CRDs if need be?

I could modify the PR to use `pkg/apis/lighthouse/v1alpha1` if you'd rather.

Agree on fixing extensibility of plugins. TBH the thing thats most pressing really is `presubmits`/`postsubmits`; its very common for repos to want their own `presubmits`/`postsubmits`; whereas having different plugins on a per repos is way less likely. e.g. I'd be happy with global plugin configurations for all the repos I work on day to day - but I really want an easy way to configure different `presubmits`/`postsubmits` in a simple way.

Maybe we should start off just focussing on `presubmits`/`postsubmits` and then later on support separate config files for plugins?

abayer:
Sounds reasonable - https://github.com/kubernetes/test-infra/blob/master/prow/config/inrepoconfig.go already has an implementation of this sort of approach.

jstrachan:
@abayer updated the PR to move the `RepositoryConfig` types into the `pkg/apis/v1alpha1` package.

as an aside looks like tekton images have vanished so any BDD test using tekton is currently failing....

abayer:
@jstrachan Yup, I mentioned that over in another PR. =)

jstrachan:
@abayer phew - green again!

abayer:
/lgtm

I'm pretty sure I'm going to want to tweak this a bunch, but it's a starting point if nothing else. =) Thanks!

jstrachan:
@abayer great stuff - agreed we’ll want to tweak. Hopefully we can support the same basic config file syntax even if code changes a fair bit

jstrachan:
@abayer I was pondering adding a PR for a minor refactor to add a little interface for getting / looking up the config/plugins given a repository in the webhook handling so we can start to create a couple of different options (eg global git repository of configs versus the current ConfigMap versus looking in the git repo that’s triggered the webhook etc). Was gonna start with a simple PR to add the interface and a default implementation that works about the same as now and see where we end up

PR #908, opened 2020-08-14
chore: Add makefile and (potentially) OWNERS to helmignore

eddycharly:
This PR adds makefile and (potentially) OWNERS to helmignore.

There should be no point in adding the Makefile in helm package.
I added OWNERS too as it is common to have an OWNERS files.

### DIFF
--- a/charts/lighthouse/.helmignore
+++ b/charts/lighthouse/.helmignore
@@ -19,3 +19,6 @@
 .project
 .idea/
 *.tmproj
+# Project files
+Makefile
+OWNERS


charts/lighthouse/.helmignore
```diff
@@ -19,3 +19,6 @@
 .project
 .idea/
 *.tmproj
+# Project files
+Makefile
```
hferentschik, line 23: Is the Makefile even used?

charts/lighthouse/.helmignore
```diff
@@ -19,3 +19,6 @@
...
+OWNERS
```
hferentschik, line 24: why adding OWNERS? There is none right?

charts/lighthouse/.helmignore
eddycharly: I don't know but it should not end up in the package.

charts/lighthouse/.helmignore
eddycharly: true, there's no OWNERS file. i added it because it's so common. i can remove it if you prefer.

PR #914, opened 2020-08-14
feat: Add helm docs

eddycharly:
This PR is a first step into getting docs for the helm chart.

This uses https://github.com/norwoodj/helm-docs to generate the docs.

It would be nice to be able to verify docs are up to date in the CI and maybe add docs gen in the makefile.

### DIFF
--- a/charts/lighthouse/Makefile
+++ b/charts/lighthouse/Makefile
@@ -1,12 +1,18 @@
 CHART_REPO := http://jenkins-x-chartmuseum:8080
 CURRENT=$(pwd)
+GO_MOD := GO111MODULE=on go
 NAME := lighthouse
 OS := $(shell uname)
 
-build: clean
+HELMDOCS := $(GOPATH)/bin/helm-docs
+$(HELMDOCS):
+	pushd /tmp; $(GO_MOD) get -u github.com/norwoodj/helm-docs/cmd/helm-docs@v0.15.0; popd
+
+build: clean $(HELMDOCS)
 	rm -rf requirements.lock
 	helm dependency build
 	helm lint
+	$(HELMDOCS)
 
 install: clean build
 	helm install . --name ${NAME}
--- a/charts/lighthouse/.helmignore
+++ b/charts/lighthouse/.helmignore
@@ -22,3 +22,4 @@
 # Project files
 Makefile
 OWNERS
+README.md.gotmpl
--- a/charts/lighthouse/Chart.yaml
+++ b/charts/lighthouse/Chart.yaml
@@ -1,5 +1,6 @@
 apiVersion: v1
-description: A Helm chart for Kubernetes
+description: |
+  This chart bootstraps installation of [Lighthouse](https://github.com/jenkins-x/lighthouse).
 icon: https://raw.githubusercontent.com/jenkins-x/jenkins-x-website/master/images/logo/jenkinsx-icon-color.svg
 name: lighthouse
 version: 0.1.0-SNAPSHOT
--- a/charts/lighthouse/README.md
+++ b/charts/lighthouse/README.md
@@ -1,1 +1,151 @@
-# golang application
+
+
+# Lighthouse
+
+This chart bootstraps installation of [Lighthouse](https://github.com/jenkins-x/lighthouse).
+
+## Installing
+
+- Add jenkins-x helm charts repo
+
+```bash
+helm repo add jenkins-x http://chartmuseum.jenkins-x.io
+
+helm repo update
+```
+
+- Install (or upgrade)
+
+```bash
+# This will install Lighthouse in the lighthouse namespace (with a my-lighthouse release name)
+
+# Helm v2
+helm upgrade --install my-lighthouse --namespace lighthouse jenkins-x/lighthouse
+# Helm v3
+helm upgrade --install my-lighthouse --namespace lighthouse jenkins-x/lighthouse
+```
+
+Look [below](#values) for the list of all available options and their corresponding description.
+
+## Uninstalling
+
+To uninstall the chart, simply delete the release.
+
+```bash
+# This will uninstall Lighthouse in the lighthouse namespace (assuming a my-lighthouse release name)
+
+# Helm v2
+helm delete --purge my-lighthouse
+# Helm v3
+helm uninstall my-lighthouse --namespace lighthouse
+```
+
+## Version
+
+Current chart version is `0.1.0-SNAPSHOT`
+
+## Values
+
+| Key | Type | Description | Default |
+|-----|------|-------------|---------|
+| `cluster.crds.create` | bool |  | `true` |
+| `clusterName` | string |  | `""` |
+| `configMaps.config` | string |  | `nil` |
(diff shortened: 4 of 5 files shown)


abayer:
Yeah, definitely let's add a makefile target and have the `pr-build` context call it - I'll see what I can do.

eddycharly:
I have some script to check docs is up to date if you want, basically it's just invoking `helm-docs` and checking that it's up to date using git...

abayer:
Yeah, if you want to give that a go, that'd be great - we should add a target that gets called as part of `check` for that in general, since we're calling the `fmt` target but not actually verifying that it didn't have to change anything.

eddycharly:
Ok, let's see what i can come up with. I added a call to `helm-docs` in the build target of the makefile.
It will probably fail as `helm-docs` is not installed. Where should i look to get it installed ?

abayer:
Actually, https://github.com/jenkins-x/lighthouse/pull/915 will do the check part. =)

abayer:
Something like https://github.com/jenkins-x/lighthouse/blob/master/Makefile#L37-L43 would be the way to install/run.

eddycharly:
Let's give a try, i should be close...

abayer:
Here's the failure:
```
GO111MODULE=off go get -u github.com/norwoodj/helm-docs/cmd/helm-docs
package k8s.io/helm/pkg/ignore: cannot find package "k8s.io/helm/pkg/ignore" in any of:
	/usr/local/go/src/k8s.io/helm/pkg/ignore (from $GOROOT)
	/home/jenkins/go/src/k8s.io/helm/pkg/ignore (from $GOPATH)
make: *** [/home/jenkins/go/bin/helm-docs] Error 1
```

Lemme try to figure out what's going on...

eddycharly:
Arf... is the a way i can view the logs ?

charts/lighthouse/Makefile
```diff
@@ -1,12 +1,18 @@
 CHART_REPO := http://jenkins-x-chartmuseum:8080
 CURRENT=$(pwd)
+GO_NOMOD := GO111MODULE=off go
 NAME := lighthouse
 OS := $(shell uname)
 
-build: clean
+HELMDOCS := $(GOPATH)/bin/helm-docs
+$(HELMDOCS):
+	$(GO_NOMOD) get -u github.com/norwoodj/helm-docs/cmd/helm-docs
```
abayer, line 9: Maybe something as dirty as adding a `git checkout -- ../../go.mod ../../go.sum` and changing this to `GO111MODULE=on go get ...` is the answer. That would fetch/build correctly and undo changes to `go.mod` and `go.sum` that would otherwise be annoying...hrrrrrrrm.

eddycharly: trying to `cd /tmp` temporarily...

abayer:
Oh riiiiight, it needs `GOMODULE=on`...and if we do that, it modifies `go.mod` and `go.sum` which we don't want it to do...

charts/lighthouse/Makefile
abayer: ...that would mess things up if there were local changes to `go.mod` and `go.sum`, though. Augh.

eddycharly:
```go
GO_MOD := GO111MODULE=on go
NAME := lighthouse	NAME := lighthouse
OS := $(shell uname)	OS := $(shell uname)


build: clean	HELMDOCS := $(GOPATH)/bin/helm-docs
$(HELMDOCS):
	cd /tmp
	$(GO_MOD) get -u github.com/norwoodj/helm-docs/cmd/helm-docs
	cd -
```

Shouldn't that do it ?

abayer: Aaaah, yeah, that'll do for now at least. We may want to end up changing to downloading a release tarball for the appropriate architecture from https://github.com/norwoodj/helm-docs/releases, exploding it, running it, and then deleting it, but the `cd /tmp` should help for now. I'm also changing #915 to do the helm steps before calling the root level `make verify-code-unchanged`...

eddycharly: is it possible to use `docker run ...` ? if so i can run it in a container instead of trying to install it

abayer: I...honestly don't know. =)

abayer:
heh, latest failure is:
```
cd -
/bin/sh: line 0: cd: OLDPWD not set
```
Try switching to `pushd /tmp; ... ; popd`?

eddycharly:
> Try switching to pushd /tmp; ... ; popd?

Let's try that

abayer:
🤞

eddycharly:
And while i'm at it, what shall i do to view the logs ? ;)

abayer:
Sadly, the logs are locked behind CloudBees authorization stuff due to Jenkins X not being quite good enough at hiding secrets in the logs. =) Just ping me when you need logs and I'll get you them, but sadly that's the best we can do for now.

abayer:
yaaaaaay it finally worked! =)

abayer:
(oh, and you can ignore the `github` etc tests failing for previous builds - the clusters they spin up for the end to end tests get garbage collected once there's a newer build's cluster for the same context/PR running, so the older build ends up failing)

eddycharly:
Thanks for your help.
I'll have to annotate the values.yaml file to get better docs gen but it can go in a follow up PR.

abayer:
I was about to override those. =)

abayer:
(sneaking in those other tests because #915 merged and so everything was gonna end up needing to retest against updated master anyway)

eddycharly:
I rebased my branch

abayer:
Ah, you didn't need to. Lighthouse just rebuilds on top of master when master's changed since the last build. =)

abayer:
Well, `pr-build` failed...for a legitish reason! The `verify-code-unchanged` target failed!

```
Code generation and/or formatting is out of date
charts/lighthouse/README.md
diff --git a/charts/lighthouse/README.md b/charts/lighthouse/README.md
index 5e7934b..0f6c1e2 100644
--- a/charts/lighthouse/README.md
+++ b/charts/lighthouse/README.md
@@ -44,8 +44,7 @@ helm uninstall my-lighthouse --namespace lighthouse
 
 Current chart version is `0.1.0-SNAPSHOT`
 
-## Chart Values
-
+## Values
 
 | Key | Type | Description | Default |
 |-----|------|-------------|---------|
@@ -149,5 +148,4 @@ Current chart version is `0.1.0-SNAPSHOT`
 | `webhooks.service.type` | string |  | `"ClusterIP"` |
 | `webhooks.terminationGracePeriodSeconds` | int |  | `180` |
 
-
 You can look directly at the [values.yaml](./values.yaml) file to look at the options and their default values.
make: *** [verify-code-unchanged] Error 1
```

Looks like the `helm-docs` output was slightly different than what's in the PR.

eddycharly:
Probably because I don’t have the same helm-docs version... will re generate docs with the latest version

eddycharly:
Should be good this time ! 🤞

eddycharly:
I pinned to version to avoid bad surprise.

PR #917, opened 2020-08-14
docs: Initiate plugins documentation

eddycharly:
This PR initiates plugins documentation.

It has the plugins list, a brief description of the plugins.yaml file structure with an example of all stanzas and their use.

It will need more work to add documentation of each plugin, but this can be done in follow up PRs.


abayer:
/lgtm

Nice idea!

PR #918, opened 2020-08-14
chore: Remove golint and blunderbuss plugins config (dead code)

eddycharly:
This PR removes `golint` and `blunderbuss` plugins config as they are not used in the code.

### DIFF
--- a/charts/lighthouse/templates/plugins-cm.yaml
+++ b/charts/lighthouse/templates/plugins-cm.yaml
@@ -22,7 +22,6 @@ data:
       # repos are specified here as:
       # - org/repo
       require_self_approval: true
-    blunderbuss: {}
     cat: {}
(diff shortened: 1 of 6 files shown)


abayer:
I actually started trying to port the golint plugin over recently but got stuck figuring out how to deal with the line by line comments in a way that would work with GitLab and BitBucket Server. I’ll probably come back to it at some point, but I don’t see a problem in removing the config for now. The only concern I could have would be that existing Prow configs still need to be valid, but I believe it’ll just ignore unknown fields when unmarshaling anyway.

abayer:
A unit test failed in `pkg/repoconfig/merge`

eddycharly:
As I’m writing plugins docs I started cleaning up the package at the same time. 

I’ll look at the failing test tonight or tomorrow.

PR #920, opened 2020-08-15
docs: Add wip plugin docs

eddycharly:
This PR adds `wip` plugin docs.


abayer:
Seems like it probably makes sense to combine these all into a single PR?

eddycharly:
I thought it would be easier to review with one pr per plugin. 

I can group them if you prefer though.

PR #927, opened 2020-08-16
feat: Add pipeline_run_params support to lighthouse job

eddycharly:
This PR adds `pipeline_run_params` support to lighthouse job.

Work in progress

### DIFF
--- a/pkg/apis/lighthouse/v1alpha1/types.go
+++ b/pkg/apis/lighthouse/v1alpha1/types.go
@@ -140,6 +140,8 @@ type LighthouseJobSpec struct {
 	// PipelineRunSpec provides the basis for running the test as a Tekton Pipeline
 	// https://github.com/tektoncd/pipeline
 	PipelineRunSpec *tektonv1beta1.PipelineRunSpec `json:"pipeline_run_spec,omitempty"`
+	// PipelineRunParams are the params used by the pipeline run
+	PipelineRunParams []config.PipelineRunParam `json:"pipeline_run_params,omitempty"`
 	// PodSpec provides the basis for running the test under a Kubernetes agent
 	PodSpec *corev1.PodSpec `json:"pod_spec,omitempty"`
 }
--- a/pkg/apis/lighthouse/v1alpha1/zz_generated.deepcopy.go
+++ b/pkg/apis/lighthouse/v1alpha1/zz_generated.deepcopy.go
@@ -5,6 +5,7 @@
 package v1alpha1
 
 import (
+	config "github.com/jenkins-x/lighthouse/pkg/config"
 	v1beta1 "github.com/tektoncd/pipeline/pkg/apis/pipeline/v1beta1"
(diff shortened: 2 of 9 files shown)


abayer:
Yeah, this does feel like a good direction to go.

eddycharly:
@abayer I adapted the tests and used the job's spec `Refs` in the templates instead of env variables.

abayer:
@eddycharly Do you want to try using this in the `test/e2e/tekton_test.go` stuff? It's using the autodetection of `git-batch-merge` currently, so I'm not sure if it's worth changing that or not.

abayer:
I'd like to keep the autodetect - it works well enough for the most common scenarios going forward, I think, so it's worth keeping.

eddycharly:
Ok, i understand. Let's keep it !

I just find it ugly that the controller has to fetch the pipeline, introspect it, and take decision on the back of the user. Plus there are chances that the task signature will change over time then it will become a nightmare... (there are already two versions of the `git-batch-merge` task).

Anyway, there is a solution to disable it if one wants to so it's not that a big issue.

eddycharly:
I made the change in the e2e tests... 🤞

pkg/apis/lighthouse/v1alpha1/types.go
```diff
@@ -140,6 +140,8 @@ type LighthouseJobSpec struct {
 	// PipelineRunSpec provides the basis for running the test as a Tekton Pipeline
 	// https://github.com/tektoncd/pipeline
 	PipelineRunSpec *tektonv1beta1.PipelineRunSpec `json:"pipeline_run_spec,omitempty"`
+	// PipelineRunParams are the params used by the pipeline run
+	PipelineRunParams []config.PipelineRunParam `json:"pipeline_run_params,omitempty"`
```
abayer, line 144: This just reminds me that I really should make it so that `config` can depend on `v1alpha1` ASAP...

eddycharly: Yep, i had to make it the other way around...
In tekton, config is part of the `apis` folder but not tied to a specific version (https://github.com/tektoncd/pipeline/tree/master/pkg/apis).

abayer: Things are uglier here because of inheriting so much from Prow, and then me trying to move more of the "config" into `LighthouseJob` and insisting on having `DeepCopy` =) I'll clean it up soon.

eddycharly: No worries, i'll help if i have time ;)

eddycharly:
Is this related to the GitHub token issue ?

abayer:
/retest

Yup, it should be ok now.

abayer:
@eddycharly There's another build of `github` running right now that should pass.

eddycharly:
I would have preferred to solve that at the pipeline run level but I failed to convince them on slack and api meeting =(

PR #938, opened 2020-08-18
feat: migrate tektoncontroller to controller-runtime

### LINKED ISSUE #933: Migrate tekton controller to controller-runtime
There’s a lot of boilerplate code in the tekton controller.

How about migrating the controller to controller-runtime ? This would simplify the implementation enormously and would be a lot easier to maintain.

eddycharly:
This PR migrates tektoncontroller to controller-runtime based reconcilier.

This is work in progress, I opened the PR early if one has suggestions/feedback.

/cc @abayer 

closes #933


abayer:
Yay! I'll try to do something like this with the foghorn controller too.

pkg/engines/tekton/utils.go
```diff
@@ -0,0 +1,256 @@
...
+const (
+	controllerName          = "tekton-controller"
+	gitCloneCatalogTaskName = "git-clone"
+	gitCloneURLParam        = "url"
+	gitCloneRevisionParam   = "revision"
+	gitMergeCatalogTaskName = "git-batch-merge"
+	gitMergeBatchRefsParam  = "batchedRefs"
+)
+
+var reProwExtraRef = regexp.MustCompile(`PROW_EXTRA_GIT_REF_(\d+)`)
```
eddycharly, line 33: @abayer do you plan on getting this magic stuff work in lighthouse ?
It looks like it's used to validate the pipeline run but is never applied.
If it's ok i can remove it :D

abayer: I'd say remove it. I never quite figured out what the point was anyway. =)

hferentschik:
What is the driver for this and what does one gain by using controller-runtime? Just trying to understand the motivation.

abayer:
Failure, btw, is lint errors on exported things that should have comments, but I assumed you weren't worrying about that yet anyway. =)

eddycharly:
@hferentschik it eliminates most of the boilerplate code used to create a controller/reconciler.

It also helps getting a clean implementation, if you look at the current implementation, it's in fact 2 listeners, that can lead to situations where both listeners fight each other. Here the reconciliation always happens based on what is observed.

In the end the code should be a lot easier to read and maintain. Seriously, the tektoncontroller is only reconciling a `LighthouseJob` and a `PipelineRun`, it should not be more than a few lines of code.

eddycharly:
I don't expect it to work yet but i'll look at the linter errors, thanks.

abayer:
Yeah, I'm all for rationalizing the controller - I derived it from copy-pasting various other controllers' bits, so making it more coherent is a good thing in my books.

pkg/engines/tekton/utils.go
eddycharly: done

abayer:
@eddycharly - gosec error this time:
```
/workspace/source/cmd/tektoncontroller/main.go,46,Errors unhandled.,LOW,HIGH,"45: 	scheme := runtime.NewScheme()
46: 	lighthousev1alpha1.AddToScheme(scheme)
47: 	pipelinev1beta1.AddToScheme(scheme)
",CWE-703
/workspace/source/cmd/tektoncontroller/main.go,47,Errors unhandled.,LOW,HIGH,"46: 	lighthousev1alpha1.AddToScheme(scheme)
47: 	pipelinev1beta1.AddToScheme(scheme)
48: 
",CWE-703
```

eddycharly:
I saw that locally, was not sure if it was going to be detected in the CI !

pkg/engines/tekton/controller.go
```diff
@@ -1,743 +1,127 @@
...
-		// Return an error here so that we requeue and retry.
+		return []string{owner.Name}
+	}); err != nil {
 		return err
 	}
 
-	spec := &origJob.Spec
-
-	// Only launch for the appropriate agent types and for triggered state
-	if origJob.Status.State == v1alpha1.TriggeredState && spec.Agent == config.TektonPipelineAgent {
```
eddycharly, line 318: @abayer jobs use only `TriggeredState` and `PendingState` ?
Looks like the state property is not updated to other state.

abayer: https://github.com/jenkins-x/lighthouse/blob/master/pkg/foghorn/controller.go#L265 - we update it in `foghorn` from the activity record on the status.

eddycharly: 👍

abayer: My reasoning for doing this in `foghorn` is to have the engine controllers (Tekton and https://github.com/jenkins-x/lighthouse-jx-controller/) do as little as possible, so their job is to handle triggering and to update the status fields that require engine-specific information/decisions (i.e., `Activity` and `ReportURL`), while `foghorn` waits for the engine controllers to make changes to those fields and then does "real" updates to the status and to the SCM provider, etc.

It may not be a great architecture, so I'm not unopposed to changing it, but it feels like it makes sense to me.

eddycharly: Yeah, i wondered what was the purpose of `ActivityRecord`. Now i know !
It probably makes sense but it's not easy to get it at first sight.

pkg/engines/tekton/controller.go
```diff
@@ -1,743 +1,136 @@
...
-			Namespace:    lj.Spec.Namespace,
-			Labels:       prLabels,
-		},
-		Spec: *specCopy,
-	}
-	// Set a default timeout of 1 day if no timeout is specified
-	if p.Spec.Timeout == nil {
-		p.Spec.Timeout = &metav1.Duration{Duration: 24 * time.Hour}
-	}
-	p.OwnerReferences = []metav1.OwnerReference{{
```
eddycharly, line 502: I need to take care here as the `Controller` property was not set. Therefore it won't be considered by the indexer.

pkg/engines/tekton/utils.go
```diff
@@ -0,0 +1,206 @@
...
+		return nil, errors.New("empty BuildID in status")
+	}
+
+	prLabels, annotations := jobutil.LabelsAndAnnotationsForJob(lj, buildID)
+	specCopy := lj.Spec.PipelineRunSpec.DeepCopy()
+	p := tektonv1beta1.PipelineRun{
+		ObjectMeta: metav1.ObjectMeta{
+			Annotations:  annotations,
+			GenerateName: fmt.Sprintf("%s-", lj.Spec.Job),
+			Namespace:    lj.Spec.Namespace,
```
eddycharly, line 57: Need to not set the namespace here, the target namespace can be different that the lighthouse job namespace.

abayer:
`tekton` failure (and presumably the others will fail too) is because no status was getting reported to GH.

eddycharly:
hmmm... what's the code responsible for that ? i might have break something.

I'll run it locally to check.

abayer:
If the current run fails, I'll take a look at the cluster.

eddycharly:
I need to polish the reconciliation loop a bit more, and i'll work on the tests.

I need to address a compatibility issue with existing jobs (those who don't have the `Controller: true` in owners reference).

abayer:
Oh, and the relevant code for updating the GitHub status etc is all in `pkg/foghorn/controller.go`. But I'm betting something useful will be in either the `foghorn` or `tekton` controller logs...but I screwed up and am not capturing the failed logs automatically right now. Fixed that over in #939, but I'll grab the logs from the next failed run.

abayer:
@eddycharly Ah, it's actually pretty blatant:

```
E0818 16:04:10.542899       1 reflector.go:153] pkg/mod/k8s.io/client-go@v0.17.6/tools/cache/reflector.go:105: Failed to list *v1beta1.TaskRun: taskruns.tekton.dev is forbidden: User "system:serviceaccount:lh-test:lighthouse-tekton-controller" cannot list resource "taskruns" in API group "tekton.dev" at the cluster scope
```

eddycharly:
Oh ! It's trying to listen to all resource types in the group instead of `PipelineRun`s only :(

Fixing right now.

eddycharly:
RBAC issue should be fixed 🤞

abayer:
@eddycharly now it's in a crash loop backoff with:

```
{"component":"lighthouse-tekton-controller","error":"no kind is registered for the type v1beta1.TaskRun in scheme \"pkg/runtime/scheme.go:101\"","file":"/workspace/source/cmd/tektoncontroller/main.go:90","func":"main.main","level":"fatal","msg":"Unable to create controller","time":"2020-08-18T16:32:43Z"}
```

=)

eddycharly:
Damn !!

abayer:
Compilation error on main.go now due to an unused import. =)

eddycharly:
That’s what you get when you change some code with phone in the metro ^^

abayer:
Still got the `no kind is registered for v1beta1.TaskRun` error...

pkg/engines/tekton/controller.go
```diff
@@ -1,743 +1,138 @@
...
-		// workqueue.
-		if key, ok = obj.(string); !ok {
-			// As the item in the workqueue is actually invalid, we call
-			// Forget here else we'd go into a loop of attempting to
-			// process a work item that is invalid.
-			c.queue.Forget(obj)
-			c.logger.Warnf("expected string in workqueue but got %#v", obj)
+// SetupWithManager sets up the reconcilier with it's manager
+func (r *LighthouseJobReconciler) SetupWithManager(mgr ctrl.Manager) error {
+	if err := mgr.GetFieldIndexer().IndexField(&pipelinev1beta1.TaskRun{}, jobOwnerKey, func(rawObj runtime.Object) []string {
```
abayer, line 46: Should this be `PipelineRun` rather than `TaskRun`?

eddycharly: Definitely, good catch !

eddycharly:
@abayer is there a makefile target to help building/pushing the docker image ?

pkg/engines/tekton/controller.go
eddycharly: That was the main reason behind the issue, thanks :D

eddycharly:
Note to myself: it's gonna be difficult to have pipeline runs and lighthouse jobs in separate namespaces because cross-namespace owner ref is not supported.

eddycharly:
Hmmm... I’ll need the logs from the tekton job please 🙏

pkg/engines/tekton/test_data/controller/start-batch-pullrequest/expected-pr.yml
```diff
@@ -17,10 +17,13 @@ metadata:
     lighthouse.jenkins-x.io/type: presubmit
   generateName: github-
   namespace: jx
+  resourceVersion: '1'
   ownerReferences:
-    - apiVersion: lighthouse.jenkins.io/v1
```
eddycharly, line 21: I don't understand how tests passed before.

pkg/engines/tekton/test_data/controller/start-pullrequest/expected-pr.yml
```diff
@@ -17,10 +17,13 @@ metadata:
     lighthouse.jenkins-x.io/type: presubmit
   generateName: github-
   namespace: jx
+  resourceVersion: '1'
   ownerReferences:
-    - apiVersion: lighthouse.jenkins.io/v1
```
eddycharly, line 21: I don't understand how tests passed before.

eddycharly:
Unit tests have been adapted to work with controller-runtime.

abayer:
The tekton controller has this in its logs:

```
{"component":"lighthouse-tekton-controller","controller":"tekton-controller","file":"/workspace/source/pkg/engines/tekton/controller.go:134","func":"github.com/jenkins-x/lighthouse/pkg/engines/tekton.(*LighthouseJobReconciler).Reconcile","level":"error","msg":"A lighthouse job should never have more than 1 pipeline run","time":"2020-08-19T11:34:14Z"}
```

Which is resulting in it never updating the `LighthouseJob` status and `foghorn` never reporting back.

abayer:
...and there are in fact two `PipelineRun`s getting created.

eddycharly:
Ah, funny, i was expecting this but didn't observe it locally. Makes sense !

abayer:
`pr-build` failure:

```
              	Spec:       v1alpha1.LighthouseJobSpec{Type: "presubmit", Agent: "tekton-pipeline", Namespace: "**", Job: "github", Refs: &v1alpha1.Refs{Org: "jenkins-x", Repo: "lighthouse", RepoLink: "https://github.com/jenkins-x/lighthouse", BaseRef: "master", BaseSHA: "e8d56b5ee9671599c75644af574a251dd3b94a5c", BaseLink: "https://github.com/jenkins-x/lighthouse/commit/e8d56b5ee9671599c75644af574a251dd3b94a5c", Pulls: []v1alpha1.Pull{{Number: 813, Author: "abayer", SHA: "dd64c739442d505cf5381e2a14b60968e8a0d86e", Link: "https://github.com/jenkins-x/lighthouse/pull/813.diff", CommitLink: "https://github.com/jenkins-x/lighthouse/pull/813/commits/dd64c739442d505cf5381e2a14b60968e8a0d86e", AuthorLink: "https://github.com/abayer"}}, CloneURI: "https://github.com/jenkins-x/lighthouse.git"}, Context: "github", RerunCommand: "/test github", PipelineRunSpec: &v1beta1.PipelineRunSpec{PipelineRef: &v1beta1.PipelineRef{Name: "jenkins-x-charts-**-build-templ-wbbx6-7", APIVersion: "tekton.dev/v1beta1"}, ServiceAccountName: "tekton-bot", PodTemplate: &pod.Template{}}, PipelineRunParams: []config.PipelineRunParam{{Name: "branch-name", ValueTemplate: "{{ range $i, $v := .Pulls }}{{if $i}} {{end}}{{ $v.SHA }}{{ end }}"}, {Name: "repo-url", ValueTemplate: "{{ .CloneURI }}"}}},
              	Status: v1alpha1.LighthouseJobStatus{
            - 		State:        "pending",
            + 		State:        "triggered",
              		ActivityName: "",
              		Description:  "",
              		... // 6 identical fields
              	},
              }
    --- FAIL: TestReconcile/start-batch-pullrequest (0.01s)
        controller_test.go:111: LighthouseJob did not match expected:   &v1alpha1.LighthouseJob{
              	TypeMeta:   v1.TypeMeta{Kind: "LighthouseJob", APIVersion: "lighthouse.jenkins.io/v1alpha1"},
              	ObjectMeta: v1.ObjectMeta{Name: "f46327af-b47e-11ea-b797-9256b7b8d9b0", Namespace: "**", ResourceVersion: "1", Labels: map[string]string{"created-by-lighthouse": "true", "lighthouse.jenkins-x.io/branch": "PR-813", "lighthouse.jenkins-x.io/buildNum": "7828158075477027098", "lighthouse.jenkins-x.io/context": "github", "lighthouse.jenkins-x.io/id": "f46327af-b47e-11ea-b797-9256b7b8d9b0", "lighthouse.jenkins-x.io/job": "github", "lighthouse.jenkins-x.io/refs.org": "jenkins-x", "lighthouse.jenkins-x.io/refs.pull": "813", "lighthouse.jenkins-x.io/refs.repo": "lighthouse", "lighthouse.jenkins-x.io/type": "presubmit"}, Annotations: map[string]string{"lighthouse.jenkins-x.io/job": "github"}},
              	Spec:       v1alpha1.LighthouseJobSpec{Type: "presubmit", Agent: "tekton-pipeline", Namespace: "**", Job: "github", Refs: &v1alpha1.Refs{Org: "jenkins-x", Repo: "lighthouse", RepoLink: "https://github.com/jenkins-x/lighthouse", BaseRef: "master", BaseSHA: "e8d56b5ee9671599c75644af574a251dd3b94a5c", BaseLink: "https://github.com/jenkins-x/lighthouse/commit/e8d56b5ee9671599c75644af574a251dd3b94a5c", Pulls: []v1alpha1.Pull{{Number: 813, Author: "abayer", SHA: "dd64c739442d505cf5381e2a14b60968e8a0d86e", Ref: "refs/pull/813/head", Link: "https://github.com/jenkins-x/lighthouse/pull/813.diff", CommitLink: "https://github.com/jenkins-x/lighthouse/pull/813/commits/dd64c739442d505cf5381e2a14b60968e8a0d86e", AuthorLink: "https://github.com/abayer"}, {Number: 814, Author: "abayer", SHA: "abcdefg", Ref: "refs/pull/814/head", Link: "https://github.com/jenkins-x/lighthouse/pull/814.diff", CommitLink: "https://github.com/jenkins-x/lighthouse/pull/814/commits/abcdefg", AuthorLink: "https://github.com/abayer"}}, CloneURI: "https://github.com/jenkins-x/lighthouse.git"}, Context: "github", RerunCommand: "/test github", PipelineRunSpec: &v1beta1.PipelineRunSpec{PipelineRef: &v1beta1.PipelineRef{Name: "jenkins-x-charts-**-build-templ-wbbx6-7", APIVersion: "tekton.dev/v1beta1"}, ServiceAccountName: "tekton-bot", PodTemplate: &pod.Template{}}, PipelineRunParams: []config.PipelineRunParam{{Name: "batch-refs", ValueTemplate: "{{ range $i, $v := .Pulls }}{{if $i}} {{end}}{{ $v.Ref }}{{ end }}"}, {Name: "branch-name", ValueTemplate: "{{ .BaseRef }}"}, {Name: "repo-url", ValueTemplate: "{{ .CloneURI }}"}}},
              	Status: v1alpha1.LighthouseJobStatus{
            - 		State:        "pending",
            + 		State:        "triggered",
              		ActivityName: "",
              		Description:  "",
              		... // 6 identical fields
              	},
              }
FAIL
```

abayer:
...but on the bright side, `tekton` looks good so far!

eddycharly:
`tekton` will eventually be green, but i'm not completely happy with the implementation, not sure how to get something really reliable

abayer:
hmm, `tekton` failed on the last part, waiting for the merge to happen after `/override`. I'm wondering if the earlier failed builds are getting re-reported...

abayer:
Yes, verified - `/override` is still kicking off a build when it shouldn't. I'll figure out why in a few minutes. =)

eddycharly:
The main issue is that it’s writing to the api server but reading from a cache. When it reads, the cache is not refreshed yet :(

Trying to find a pattern but it’s not that simple.

abayer:
Ah, right, so we're not filtering the `LighthouseJob`s to only trigger for ones that have the `tekton-pipelines` agent and have their state as triggered, and `/override` ends up creating a job with a state of success.

eddycharly:
Good catch again, I forgot to filter on the agent 🙃
This one is easy !

eddycharly:
Best pattern I can think about is ignore status changes on the job, set the pending status when the pipeline run is created... that should prevent invalid cache reads.

abayer:
@eddycharly - don't just filter on the agent, also filter on the `Status.State` - we only want to kick off when the state is `Triggered`...

eddycharly:
I fear that it won't work because of the cache.

Ideally, I feel like the `Triggered` to `Pending` state transition should be consequence of getting the `PipellineRun` created and should not be set manually.

abayer:
Perhaps - but for now we do need to continue the existing behavior, since https://github.com/jenkins-x/lighthouse-jx-controller/ is working that way and we need to be consistent, unless we update that too. This is also the Prow behavior, fwiw.

eddycharly:
I kept the current behavior, it could be improved later, i'll ask the question on controller-runtime github or slack what's the best approach in our case.

As controller-runtime supports leader election, it could be useful to test with multiple replicas.

eddycharly:
Those controllers look very similar.
Why is the lighthouse-jx-controller in another repo ?

eddycharly:
Okay ! `tekton` job is green 🎉

abayer:
Yaaaaay!

eddycharly:
The last issue i see is about the owner ref not containing the `Controller: true` property. How do you want to deal with that ?

We can just ignore it but jobs running won't get their activity record updated.
Or we can relax the checks in the indexer.

Only the running jobs are an issue here. What do you think ?

abayer:
I'm fine with things not quite working right for jobs that are running at upgrade time - there aren't enough people using Lighthouse + Tekton in production at this point for it to be a major problem, so let's go ahead and do it right going forward without excessive complexity for backwards compat.

pkg/engines/tekton/controller.go
eddycharly: > I'm fine with things not quite working right for jobs that are running at upgrade time - there aren't enough people using Lighthouse + Tekton in production at this point for it to be a major problem, so let's go ahead and do it right going forward without excessive complexity for backwards compat.

👍

eddycharly:
Looks like everything is green, it should be ready for review.

PR #944, opened 2020-08-19
fix: Speed up codegen considerably

abayer:
Like from 5m37s to 7s on my laptop. Turning off `GO111MODULE` makes for a huge improvement - see https://github.com/kubernetes/code-generator/issues/69


abayer:
Dangit, doesn't work in CI but works for me locally. Time to experiment.

abayer:
Ok, think I got it working - my second attempt failed because it didn't delete the temporary vendor directory, but now `update-codegen.sh` should do that properly.

abayer:
Yaaaaay that does the trick.

/assign @hferentschik

PR #949, opened 2020-08-20
fix: Don't fallback to master as `git-clone` parameter for postsubmits

### LINKED ISSUE #948: Tekton rev to clone for postsubmit ends up as "master" always
Which isn't right.

abayer:
We have access to the branch name in these cases so let's use that.

fixes #948


pkg/engines/tekton/utils.go
```diff
@@ -111,7 +111,7 @@ func makePipelineRun(ctx context.Context, lj v1alpha1.LighthouseJob, namespace s
 				if len(lj.Spec.Refs.Pulls) > 0 {
 					env[paramNames.revParam] = lj.Spec.Refs.Pulls[0].SHA
 				} else {
-					env[paramNames.revParam] = "master"
+					env[paramNames.revParam] = lj.Spec.Refs.BaseRef
```
hferentschik, line 114: Want to add some tests? Or is there no good/easy way to test this?

PR #951, opened 2020-08-20
WIP: feat: Add ability to autoincrement build numbers

### LINKED ISSUE #950: Add ability to generate sequential, autoincrementing build IDs/numbers
This isn't relevant for Jenkins X, since that handles build numbers internally, but it'd be nice to generate that dynamically/uniquely (for some value of "unique") for Tekton. Prow's `tot` utilizes the expected location of build logs to fallback on when it can't find a name in its in-container JSON, but I'd rather go with a config map and not worry about the fallback.

abayer:
Currently available only for Tekton. It'll start at 1 for each combination of `(org)-(repo)-(branch or PR)-(context)`. The "history" is persisted in a configmap - there may be a better option/approach.

This is heavily derived from Prow's `tot` service.

Still needs to be worked into the e2e tests.

fixes #950

### DIFF
--- a/Dockerfile.buildnumbers
+++ b/Dockerfile.buildnumbers
@@ -1,0 +1,4 @@
+FROM alpine:3.10
+RUN apk add --update --no-cache ca-certificates git
+COPY ./bin/buildnumbers  /buildnumbers
+ENTRYPOINT ["/buildnumbers"]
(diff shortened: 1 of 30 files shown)


eddycharly:
Why not define a CRD for that ?
It would be much simpler.

PR #961, opened 2020-08-21
chore: remove deck and sinker config (deadcode)

eddycharly:
This PR removes deck and sinker config (deadcode).


abayer:
Yeah, that's there to try to preserve the existing templates for report comments, but I'm fine with changing where that's recorded.

eddycharly:
@abayer Ah yes, it's responsible of this line in the report comment:
image

The table and details come from somewhere else.

I'm going to check what can be cleaned.

PR #966, opened 2020-08-24
chore: remove Gerrit config

eddycharly:
This PR removes Gerrit config


shreelola:
Why was Gerrit removed?

PR #967, opened 2020-08-24
chore: Switch Keeper to config map watcher

abayer:
This'll keep it from failing to start if the configmaps aren't present and/or are malformed, and avoids the need for mounting the configmap as well.

And I fixed up the out-of-the-box `config` configmap to not be invalid.


hferentschik:
/lgtm

On a side note, what's about creating issues? What's the approach we want to adopt there?

abayer:
Yeah, I probably should have created one. Sorry. =)

hferentschik:
> Yeah, I probably should have created one. Sorry. =)

no worries :-)

Issue #965, opened 2020-08-24
Rework config package


eddycharly:
How about splitting the config package per concern ?

I find it a bit hard to follow, things are not really organised and most of the structs are mixed together. 

One concern that could be placed in a separate package could be everything related to jobs ?

eddycharly:
Job config concern was addressed in #968 .

Next is Lighthouse config concern.

eddycharly:
One last thing, shall we move the config package into apis ?
It looks to me that the config is part of the api.

abayer:
I'm debating that - for now, I'd say no...

eddycharly:
Ok, can close the issue.

PR #968, opened 2020-08-24
refactor: rework jobs config

eddycharly:
This PR moves all job related config into its own package.

This is work in progress, i feel like there some improvements that can be done around initialization/validation.

Related issue #965


abayer:
fwiw, lint failures:
```
pkg/config/job/base.go:25:6: type name will be used as job.JobBase by other packages, and that stutters; consider calling this Base
pkg/config/job/brancher.go:45:1: exported method Brancher.SetBrancherRegexes should have comment or be unexported
pkg/config/job/regexpchangematcher.go:61:1: exported method RegexpChangeMatcher.SetChangeRegexes should have comment or be unexported
```

eddycharly:
Yeah, I didn’t focus on linting yet.
Moving code around, fixing tests, identifying some strange constructs that I would like to implement differently...

eddycharly:
I was thinking about using dot imports but not sure if it’s good practice or not...

eddycharly:
It should be ready for review. There are probably a few things that could be improved but most of the work has been done.

I will continue with lighthouse config related concern now.

Issue #885, opened 2020-08-04
Support Jenkins job execution engine


abayer:
This would be a first part - just replicating the Prow-style Jenkins integration. But longer-term, I'd like to see something more intricate, with integration with Jenkins multibranch to reuse jobs created by Jenkins automatically.

PR #970, opened 2020-08-24
WIP: feat: Add AdditionalFields map to LighthouseJobSpec and JobBase

### LINKED ISSUE #969: Enable agent engine implementations to have their own job config fields
Right now, we have to add any agent engine-specific field (like `PipelineRunSpec`) to `LighthouseJobSpec` and `JobBase`, which is not ideal at all. It should be possible for agent engine implementations to have their own arbitrary job configuration fields without requiring changes to `LighthouseJobSpec` and `JobBase`, so that we can actually support agent engines as "plugins".

abayer:
This is the best option I've got currently for a way for agent engine implementations to have their own configuration fields without requiring that we have those fields baked into `LighthouseJobSpec` and `JobBase`. I'm open to other approaches.

fixes #969

### DIFF
--- a/docs/config/Lighthouse config.md
+++ b/docs/config/Lighthouse config.md
@@ -172,6 +172,7 @@ JobBase contains attributes common to all job types
 | Spec | `spec` | *v1.PodSpec | No | Spec is the Kubernetes pod spec used if Agent is kubernetes. |
 | PipelineRunSpec | `pipeline_run_spec` | *tektonv1beta1.PipelineRunSpec | No | PipelineRunSpec is the Tekton PipelineRun spec used if agent is tekton-pipeline |
 | PipelineRunParams | `pipeline_run_params` | [][PipelineRunParam](#PipelineRunParam) | No | PipelineRunParams are the params used by the pipeline run |
+| AdditionalFields | `additional_fields` | map[string]string | No | AdditionalFields is a string/string map for specifying additional fields. This can be used by agent implementations<br />for configuration that's not baked into the spec. |
 |  |  | [UtilityConfig](#UtilityConfig) | Yes |  |
 
 ## JobConfig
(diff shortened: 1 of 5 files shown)


eddycharly:
Do you have a specific use case in mind ?

abayer:
@eddycharly - What's directly prompted me is thinking about Jenkins triggering support (i.e., #885) and wanting to be able to have job configuration specific for Jenkins without having to hardcode it in the structs.

eddycharly:
Yep but it's not typed anymore, then you can't say if the configuration is actually for the engine.
Having a typed structure looks like a good thing to me.

abayer:
How can we do a typed structure where the typed structure is defined outside of this repo, though, without getting into circular dependency hell?

eddycharly:
Not sure i understand correctly, if it's a reference to an external package, there's probably no issue (we do it already with tekton).

Do you have an example of jenkins configuration that would go in here ?

abayer:
Lemme see if I can explain better...

So `lighthouse-jx-controller` is in a separate repository (for a number of reasons, most notably getting rid of Lighthouse's direct dependency on Jenkins X code). While some engine controllers may always live in the main Lighthouse repository, I think we also want to make it possible for someone to write an engine controller without having to make changes in lighthouse.git to support it, and without lighthouse.git depending on that engine controller's repository. That would mean a few things:
* Agent type strings would have to be discoverable, rather than hardcoded in Lighthouse code (presumably we'd add a CRD for engines).
* That same CRD would also have engine-specific but non-job configuration if needed - global configuration for all jobs using that engine, basically.
* Any job configuration specific to that engine (like `PipelineRunSpec` for the Tekton controller, for example) couldn't be defined in the `job.Base` struct or the `LighthouseJobSpec` struct - because Lighthouse proper wouldn't know about this new engine.

The third point is the one I'm trying to address here. As I said, I'm very much open to other ways to solve it - and it may be the case that we can't make it possible to define an engine, its controller, and its job configuration entirely outside of lighthouse.git. I'd like to try to find a way to do that, but if we can't find a good way, then c'est la vie. =)

eddycharly:
Ah yes... i see, it would be great but sounds difficult indeed.

Maybe tekton has some opinion on this, it looks somewhat similar to custom `Task`s.

abayer:
Yeah, I'm gonna put this away for now.

eddycharly:
I think i have some ideas but i am not sure if this is worst the troubles as it would increase complexity.
I guess that LighthouseJobs were not built with this in mind.

abayer:
Yup.

eddycharly:
I feel like external engines would work with their own CRD.
A LighthouseJob would then exist to bootstrap the dedicated resource creation and provide the necessary primitives for engine implementors to make the status evolve ?

abayer:
Yeah, I think that may make sense. Let's talk about this more over on #969.

PR #972, opened 2020-08-25
fix: allow to run controllers locally

### LINKED ISSUE #971: Unable to run controllers locally for development and debugging purposes
It used to be possible to run the controllers locally for development. This was achieved by trying to resepct the environment variable _KUBE_CONFIG_ resp. checking the path `~/.kube/config` for a valid configuration file:

```go
func GetConfig(masterURL, kubeconfig string) (*rest.Config, error) {
	if kubeconfig == "" {
		kubeconfig = os.Getenv("KUBECONFIG")
	}
	// If we have an explicit indication of where the kubernetes config lives, read that.
	if kubeconfig != "" {
		return clientcmd.BuildConfigFromFlags(masterURL, kubeconfig)
	}
	// If not, try the in-cluster config.
	if c, err := rest.InClusterConfig(); err == nil {
		return c, nil
	}
	// If no in-cluster config, try the default location in the user's home directory.
	if usr, err := user.Current(); err == nil {
		if c, err := clientcmd.BuildConfigFromFlags("", filepath.Join(usr.HomeDir, ".kube", "config")); err == nil {
			return c, nil
		}
	}

	return nil, fmt.Errorf("could not create a valid kubeconfig")
}
```

This functionality was removed as part of 85af14892f38c1f251d439a8563bb9e585daf8dc in an attempt to simplify the code. 

It should be resurrected, since being able to run the controllers locally and in particular debug them locally is extremely useful.

hferentschik:
fix: allow to run controllers locally

This reverts commit bb2e6479
fixes #971

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ GO_LDFLAGS :=  -X $(PROJECT)/pkg/version.Version='$(VERSION)'
 GO_DEPENDENCIES := $(call rwildcard,pkg/,*.go) $(call rwildcard,cmd/,*.go)
 
 .PHONY: all
-all: docs build test check
+all: build test check docs
 
 .PHONY: test
 test: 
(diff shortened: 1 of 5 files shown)


eddycharly:
Ooops, sorry for that !
I thought it would fallback to the `KUBECONFIG` env var but it looks like no.

PR #974, opened 2020-08-25
refactor: rework lighthouse config

eddycharly:
This PR moves all lighthouse config to its own package.

Related issue #965


eddycharly:
This was a big one !

PR #975, opened 2020-08-25
chore: removing obsolete files and directories

### LINKED ISSUE #973: Remove obsolete files
There are several files in the repository which don't fulfil any function given the current setup (build, release, etc):

* `skaffold.yaml` and `watch.yaml` - Skaffold related files, but not working in this context. Probably a left-over from a Jenkins X Go quickstart 
* `config.yaml`
* `dependency-matrix` 
* `charts/preview` - not used anymore
* `hack/boilerplate/boilerplate.go.txt` - empty and unused

hferentschik:
fixes #973


abayer:
We can remove `config.yaml` and `custom-boilerplate.go.txt` as well - I forgot to remove the latter when I switched the codegen around. Oops. =)

abayer:
And while we're at it, might as well remove `charts/preview` since that hasn't been used in ages.

hferentschik:
> And while we're at it, might as well remove charts/preview since that hasn't been used in ages.

Will do

hferentschik:
> We can remove config.yaml and custom-boilerplate.go.txt as well - I forgot to remove the latter when I switched the codegen around

Jupp. Even though I think I tried to remove the boilerplate once and it was complaining about something. However, I think that was just something which needed adjusting. I'll take care of this.

PR #987, opened 2020-08-27
fix: Use proper v1 format for CRD

### LINKED ISSUE #986: Update CRD to apiextensions.k8s.io/v1
We've already got that in `crds/lighthouse.jenkins.io_lighthousejobs.yaml`, generated via `make crd-manifests` (sorta), but `charts/lighthouse/config/lighthousejobs.lighthouse.jenkins.io.yaml`, the one actually used in the chart. And `make crd-manifests` will actually fail currently because Tekton Pipelines v0.14.2 has a field nested down within `PipelineRunSpec` that doesn't have a JSON tag.

So! Let's...
* Unify on one copy of the CRD YAML in the repo, rather than two.
* Generate that CRD YAML via `make crd-manifests` without it failing, which means in practice upgrading our dependency to Tekton Pipelines v0.15.2.
* Add `make crd-manifests` to our `pr-build` context to make sure it's consistent.

abayer:
Which involved upgrading our Tekton dependency and symlinking the generated CRD. I'm not sure if it'd be better to read the `crds/...` file directly instead?

fixes #986


eddycharly:
I don't think it will work, the generated CRD is not valid IIRC.

Some structs are using custom json marshalling and kubebuilder does not support that.

abayer:
Ah, well, let's see...it was worth trying.

eddycharly:
Yeah, i opened a PR in the controller-tools repo to workaround this limitation but no reply yet and i'm not particularly happy with the proposed solution.

I thought about a second approach, much cleaner IMO, but i would like to get some feedback from the kubebuilder guys before diving working it.

abayer:
Yeah, this is a non-starter for now the more I think about it, since Lighthouse still needs to be installable on older k8s versions anyway. So I'll just close this for now. =)

PR #988, opened 2020-08-27
chore: make webhooks ingress part of webhooks section in chart

eddycharly:
This PR makes webhooks ingress part of the webhooks section in chart.

This improves readability and allows the end user to specify the host he wants (works better with ngrok for example).


charts/lighthouse/templates/webhooks-ingress.yaml
```diff
@@ -0,0 +1,18 @@
+{{- if .Values.webhooks.ingress.enabled }}
+apiVersion: extensions/v1beta1
+kind: Ingress
+metadata:
+  name: {{ template "webhooks.name" . }}
```
abayer, line 5: For legacy and Prow compatibility reasons, the ingress name still needs to be `hook`. That's why https://github.com/jenkins-x/lighthouse/blob/master/test/e2e/helpers.go#L347 failed in the e2e tests.

charts/lighthouse/values.yaml
```diff
@@ -92,7 +92,6 @@ webhooks:
 
   # webhooks.service -- Service settings for the webhooks controller
   service:
-    name: hook
```
abayer, line 95: Still need to keep this for compat reasons. =)

eddycharly: I’ll add an override mechanism and state in the docs that it’s there for compact reasons (a regular user should not need it).

eddycharly: Both the name of the service and ingress matter ?

abayer: Regrettably, yes - Jenkins X makes some hard assumptions on both.

eddycharly: Ok, i adapted to chart templates but i think i'll need to change some config for the tests too.

abayer: Actually, the ingress name can be changed - I forgot that jx has its own hook ingress. But yeah, the Tekton tests will need to change if the ingress name changes.

eddycharly: We could look up the ingress by labels instead of relying on an hard coded name.

abayer:
The Jenkins X contexts are failing because they expect the service to be named hook.

abayer:
It's annoying but hard-coded in https://github.com/jenkins-x-charts/jxboot-resources/blob/master/jxboot-resources/templates/700-hook-ing.yaml#L20

eddycharly:
It should be named `hook` (i added `.Values.webhooks.serviceName`)

eddycharly:
Ah no, there's no `values.yaml` files in those tests...

abayer:
And we can't assume there would be, so at least for now, it has to default to `hook`.

eddycharly:
Actually it looks like there is one in `bdd/values.yaml.template`.
But some keys don't look correct (tide ?).

abayer:
That is rather old, but more importantly, it's meant to match up with https://github.com/jenkins-x/jenkins-x-boot-config/blob/master/env/lighthouse/values.tmpl.yaml as closely as possible, with the only changes that really matter being the `$VERSION` stuff. A Jenkins X install is going to make assumptions about service names, etc, so we really do need to keep it as `hook` for the foreseeable future.

eddycharly:
Alright, i've set the default to be `hook` and changed the doc:
```
  # webhooks.serviceName -- Allows overriding the service name, this is here for compatibility reasons, regular users should clear this out
  serviceName: hook
```

abayer:
Someday, once the Jenkins X release process is unborked (things are just..busted there at the moment), I'll update it to set `webhooks.serviceName` to `hook` so we can get rid of the explicit naming here.

Issue #990, opened 2020-08-27
Be able to move to building/releasing outside of a Jenkins X context


abayer:
Eventually we're going to want to build and release outside of the Jenkins X cluster we currently use, so we need to come up with pipelines etc for that.

PR #991, opened 2020-08-27
WIP: feat: Working towards build/release outside of jx

abayer:
relates to #990, but this isn't happening any time soon.

For now I'm just trying to make sure the `Pipeline`s I'm coming up with make sense.

e2e clusters will still get deleted by https://github.com/jenkins-x/jx/blob/master/pkg/cmd/step/e2e/step_e2e_gc.go or something with the same behavior - we'll need to make sure the cluster names are `pr-(number)-(context)-(monotonically increasing number)`, so we probably want to end up with `BUILD_ID` being milliseconds since epoch or something. I'll probably have a non-WIP PR up for that to replace #951 shortly.

I haven't figured out how we want to deal with capturing build logs and artifacts. Re artifacts, for the moment, I've got a `gsutil` call in `bdd/capture-failed-pod-logs.sh` for saving the various controller logs from failed runs, but I intend to go dig into how Tekton's doing this in dogfooding.

EDIT: They're not really doing anything like that in dogfooding. =) Need to think about that more.

### DIFF
--- a/charts/lighthouse/Makefile
+++ b/charts/lighthouse/Makefile
@@ -30,12 +30,22 @@ clean:
 .PHONY: snapshot
 snapshot: update-version release
 
+# TODO: Gotta figure out the right way to get the git creds in here securely
 release: clean
 	helm dependency build
 	helm lint
 	helm init --client-only
 	helm package .
-	@curl --fail -u $(CHARTMUSEUM_USER):$(CHARTMUSEUM_PASS) --data-binary "@$(NAME)-$(shell sed -n 's/^version: //p' Chart.yaml).tgz" $(CHART_REPO)/api/charts
+	git clone https://github.com/lighthouse-chatops/charts.git
+	pushd charts
+	git checkout gh-pages
+	mv ../"@$(NAME)-$(shell sed -n 's/^version: //p' Chart.yaml).tgz" lighthouse
+	helm repo index . --url https://lighthouse-chatops.github.com/charts
+	git add -i
+	git commit -av
+	git push origin gh-pages
+	popd
+	rm -rf charts
 	rm -rf ${NAME}*.tgz%
 
 update-version:
@@ -57,4 +67,13 @@ tag: update-version
 	git push origin v$(VERSION)
 
 delete-from-chartmuseum:
-	@curl --fail -u $(CHARTMUSEUM_USER):$(CHARTMUSEUM_PASS) -X DELETE $(CHART_REPO)/api/charts/$(NAME)/$(VERSION)
+	git clone https://github.com/lighthouse-chatops/charts.git
+	pushd charts
+	git checkout gh-pages
+	rm lighthouse/"@$(NAME)-$(shell sed -n 's/^version: //p' Chart.yaml).tgz"
+	helm repo index . --url https://lighthouse-chatops.github.com/charts
+	git add -i
+	git commit -av
+	git push origin gh-pages
+	popd
+	rm -rf charts
--- a/tekton/pipelines/lh-release-pipeline.yaml
+++ b/tekton/pipelines/lh-release-pipeline.yaml
@@ -1,0 +1,136 @@
+apiVersion: tekton.dev/v1beta1
+kind: Pipeline
+metadata:
+  name: lh-release-pipeline
+spec:
+  params:
+    - name: version
+      type: string
+      description: The version to release, without a leading "v"
+    - name: previous-version
+      type: string
+      description: The previous version, without a leading "v"
+    - name: revision
+      type: string
+      description: The revision to tag and release
+  workspaces:
+    - name: source
+  tasks:
+    - name: fetch-repo
+      taskRef:
+        name: git-clone
+      workspaces:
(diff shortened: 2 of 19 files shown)


abayer:
@eddycharly @hferentschik - I’d appreciate your thoughts on the e2e pipeline and tasks I’ve got at https://github.com/jenkins-x/lighthouse/pull/991/files#diff-03d71fbc775b7bd1c86531e273667ad8. I’m going to try to put together a first attempt at the release pipeline today, with the idea that we’ll stop releasing on every push to master, and instead release in response to tags getting pushed.

pkg/apis/lighthouse/v1alpha1/types.go
```diff
@@ -47,6 +47,8 @@ const (
 	JobNameEnv = "JOB_NAME"
 	// JobTypeEnv is the type of job
 	JobTypeEnv = "JOB_TYPE"
+	// ContextNameEnv is the job's context
+	ContextNameEnv = "CONTEXT_NAME"
```
eddycharly, line 51: Do we really need to change the CRD ?

abayer: It’s not actually changing the CRD, just adding the context key...which I guess doesn’t need to be here. I originally was going to add it to the `GetEnvVars` function but changed my mind, so yeah, I’ll move this.

eddycharly:
The pipeline looks pretty good. I don't see anything related to the release itself though.

But for build/push/tests it should do it !

eddycharly:
Also, golint/govet/verify-codegen is not there yet ?

abayer:
Yeah, I’m working on the release and unit test/lint/etc pipelines today.

eddycharly:
And for the logs, i would suggest some standard tools (fluentbit/fluentd) rather than something custom.
Personally i like https://github.com/banzaicloud/logging-operator for this.

abayer:
Actually, the more I think about it, the more I realize that triggering releases automatically from tags may not make sense - we need to know the appropriate _previous_ release tag to generate a changelog properly. Tekton releases, for example, are triggered manually (https://github.com/tektoncd/pipeline/blob/master/tekton/release-cheat-sheet.md, though that uses the stale approach of passing resources around while we'd just call a variant of what I've got in the PR e2e pipeline to build the binaries, images, and helm chart).

abayer:
Ok, updated - split things up into `tekton/tasks` and `tekton/pipelines` directories with individual files for each `Task` and `Pipeline`, and an initial `README` as well. There are now four `Pipeline`s:
* `lh-pr-e2e-pipeline` - which will be used for the various e2e contexts
* `lh-pr-checks-pipeline` - for lint/vet/sec/test
* `lh-release-pipeline` - for manually kicking off to actually perform releases
* `lh-update-tekton-resources-pipeline` - this will be set up as a postsubmit with `run_if_changed: '^tekton/(pipelines|tasks)/.*$'`, to automatically update the resources in the cluster when a PR changing them merges.

abayer:
fyi, all the jx-based e2e tests are gonna fail because I changed the chart makefile to show what it would look like for using GitHub Pages as the helm repo.

hferentschik:
@abayer - what's the state on all this? I guess it is still on the back-burner for now, right?

hferentschik:
> Actually, the more I think about it, the more I realize that triggering releases automatically from tags may not make sense 

Why not? I guess there are always options. I think releasing on a tag being pushed can work provided you don't need/want to change any repo files as part of the release. 

> we need to know the appropriate previous release tag to generate a changelog properly.

That can be worked out, right? If the changelog generation is the only thing which causes problems I would think we can find a solution for that.

> Tekton releases, for example, are triggered manually

Which is not the worst imo :-)

charts/lighthouse/Makefile
```diff
@@ -30,12 +30,22 @@ clean:
...
 snapshot: update-version release
 
+# TODO: Gotta figure out the right way to get the git creds in here securely
 release: clean
 	helm dependency build
 	helm lint
 	helm init --client-only
 	helm package .
-	@curl --fail -u $(CHARTMUSEUM_USER):$(CHARTMUSEUM_PASS) --data-binary "@$(NAME)-$(shell sed -n 's/^version: //p' Chart.yaml).tgz" $(CHART_REPO)/api/charts
+	git clone https://github.com/lighthouse-chatops/charts.git
```
hferentschik, line 39: are these dummy/test repos you created?

tekton/pipelines/lh-release-pipeline.yaml
```diff
@@ -0,0 +1,136 @@
...
+          value: $(params.revision)
+        - name: subdirectory
+          value: ""
+        - name: depth
+          value: "0"
+    - name: build-binaries
+      taskRef:
+        name: lh-build-binaries
+      runAfter:
+        - fetch-repo
```
hferentschik, line 38: not really specific to this PR, but is this really required? IMO it should be the default. I'd argue in 99% of the cases I am writing my pipeline in a way that the steps are already in sequence.

Issue #989, opened 2020-08-27
Add periodic jobs support


eddycharly:
Shall we add support for periodic jobs ?

A simple implementation could just watch the config map and keep a list of cronjobs up to date.

PR #992, opened 2020-08-27
WIP feat: add support for periodic jobs

eddycharly:
This PR adds support for periodic jobs.

Related issue https://github.com/jenkins-x/lighthouse/issues/989

Every periodic job is represented by a `CronJob` in the cluster, the `CronJob` being responsible of creating the corresponding `LighthouseJob`.

There is an additional `CronJob` that takes the responsibility of keeping the `CronJob`s list in sync with the periodic jobs list declared in the config. The scheduling of this master `CronJob` can be customized in the helm chart values.

For the record, Prow uses a custom component (`horologium`) that does all the scheduling by hand. This component cannot be scaled, is kicks off every minute (the process in not event based).

The implementation here is a lot simpler and doesn't sacrifice too much, the only thing we loose compared to Prow is that `horologium` supports standard cron expressions like `@hourly ...` etc.

/cc @abayer 

WIP: tests are missing


cmd/scheduler/main.go
```diff
@@ -0,0 +1,192 @@
...
+				APIVersion:         "batch/v1beta1",
+				Kind:               "CronJob",
+				Name:               periodic.Name,
+				Controller:         &true,
+				BlockOwnerDeletion: &true,
+			}},
+		},
+		Spec: jobutil.PeriodicSpec(periodic),
+	}
+	b, err := yaml.Marshal(lhjob)
```
eddycharly, line 146: The lh job will probably need to be serialized into json so that it's a single line.

cmd/scheduler/main.go
```diff
@@ -0,0 +1,192 @@
...
+				Spec: batchv1.JobSpec{
+					Template: corev1.PodTemplateSpec{
+						Spec: corev1.PodSpec{
+							Containers: []corev1.Container{{
+								Name:  "mklhjob",
+								Image: "bitnami/kubectl",
+								Command: []string{
+									"bash",
+									"-c",
+									"echo $LHJOB | kubectl create -f -",
```
eddycharly, line 177: We need a service account that can create lh jobs here.

eddycharly:
We would need a more marine sounding than `scheduler` too =)

abayer:
Might be worth taking a look at https://github.com/kubernetes/test-infra/blob/master/prow/cmd/horologium/main.go as a starting point, since that's the Prow equivalent.

eddycharly:
Tbh i never understood completely why prow didn't use cronjobs.

eddycharly:
The only difference i can see is that `horologium` supports more standard cron expressions.

But i'm not sure it is worth it, this complicates the implementation quite a bit and makes the whole solution less reliable IMO.

eddycharly:
FWIW, I tested it locally and it works as expected.

eddycharly:
I've been discussing this a little on prow's slack, there is no evident reason why `horologium` is better.

IMHO horologium has some serious drawbacks and i'm in favor for something simple, exploiting the built in CronJobs functionnality.

(I will remove the WIP but i'll be happy to discuss it further)

abayer:
Sounds reasonable. I’d like @hferentschik to take a look too. =)

eddycharly:
How about a better name ? I feel like it’s a bit like waves that repeat the same thing continuously... ;)

abayer:
`tides` would be perfect except for already being a thing in Prow. =)

eddycharly:
swell or swirl ?

abayer:
So a couple things from my latest look at the code:

* Definitely needs some tests - unit and ideally something that exercises the cronjob to make sure it actually is doing what it’s supposed to.
* I feel a bit weird at the cronjob kicking off `kubectl` like that...it just feels odd to me. That and I think I’d prefer if we weren’t using an image from Docker Hub in the cronjob given the rate limit changes coming up in November!

abayer:
And maybe [strobe](https://en.wikipedia.org/wiki/Lighthouse#Lens) since they’re used in some lighthouses and are "used to produce regular flashes of light"?

eddycharly:
> Definitely needs some tests - unit and ideally something that exercises the cronjob to make sure it actually is doing what it’s supposed to.

I agree, i'll add some unit tests at least.

> I feel a bit weird at the cronjob kicking off kubectl like that...it just feels odd to me. That and I think I’d prefer if we weren’t using an image from Docker Hub in the cronjob given the rate limit changes coming up in November!

I agree too, i have another approach in mind that might be better. The binary could support two commands `sync` and `start`.
- `sync` would do what the binary currently does except replacing `kubectl` in the cron job by the same binary with a call to the `start` command and removing the serialization of the lh job
- `start` would take a job name in argument, look it up in the config, create the corresponding lh job from it, and create the lh job using a regular lh client

It would eliminate the need for `kubectl` and the serialization step that i'm not particularly happy with.

eddycharly:
I like `strobe` !

abayer:
Yeah, I think I like that approach better.

eddycharly:
I switched to the second approach (and renamed everything to strobe :-)

eddycharly:
What's up with the tekton build ?

abayer:
```
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: ValidationError(CronJob.spec.jobTemplate.spec.template.spec): unknown field "strobeName" in io.k8s.api.core.v1.PodSpec
```
Which looks to be because of https://github.com/jenkins-x/lighthouse/pull/992/files#diff-b0786322408ab69304dc367fe069a30eR41 - I think you accidentally replaced `schedulerName: default-scheduler` with `strobeName: default-strobe`. =)

eddycharly:
ah, yes thanks.
fixed it 🤞

PR #995, opened 2020-08-28
chore: make e2e tests non sensible to the ingress name

eddycharly:
This PR makes e2e tests non sensible to the ingress name.

The ingress is looked up by labels rather than by name.


eddycharly:
There might be something wrong...

eddycharly:
The jsonpath expression was wrong.

abayer:
/hold

Holding briefly for getting a jx PR downstream of Lighthouse to actually merge. Will unhold soon.

PR #1004, opened 2020-09-03
fix: support in repo configuration of triggers (presubmit/postsubmits)

jstrachan:
so this works kinda like github actions where the pipelines (PipelineRun)
and Presubmit/Postsubmits are all defined in the underlying git repository
and we use that data to trigger pipelines

here's an example source git repository layout:
https://github.com/jstrachan/lighthouse/tree/local-config-plugin/pkg/webhook/test_data/myorg/myrepo

so the `.lighthouse/*/triggers.yaml` files can define presubmit/post submits locally like this [.lighthouse/linter/triggers.yaml](https://github.com/jstrachan/lighthouse/blob/local-config-plugin/pkg/webhook/test_data/myorg/myrepo/.lighthouse/linter/triggers.yaml) which can reference local `sourcePath` files like this [lint.yaml](https://github.com/jstrachan/lighthouse/blob/local-config-plugin/pkg/webhook/test_data/myorg/myrepo/.lighthouse/linter/lint.yaml)


eddycharly:
Thanks @jstrachan for working on this !

Why the plugin ? The in repo jobs are not started by the `trigger` plugin ?

For opt-in, did you consider something like what Prow did ? Make the opt-in part of the config instead ?

Also, how does the config different than `presubmits` or `postsubmits` apply here ?
For example `branchProtection` (https://github.com/jstrachan/lighthouse/blob/cc8a10914152dd04795429d3a3fa54868feb05d6/pkg/plugins/local-config-trigger/test_data/myorg/myrepo/.lighthouse/linter/lighthouse.yaml#L12-L15) ?

jstrachan:
@eddycharly currently lighthouse loads the entire config for every git repository + the presubmit/postsubmits from a global location (e.g. a single ConfigMap). What this plugin does is allow the presubmit/postsubmits to be loaded from the actual git repository. So folks can pick which repositories own their 'lighthouse configuration' - rather like how github actions works, you put the github actions configuration files inside each git repository - not in some monster huge YAML file in a central place.

BTW `presubmit/postsubmit` are configurations for the `trigger` plugin to decide when/how to trigger pipelines which is different from branch protection. e.g. you can define 3 pipelines on a `presubmit` to run 3 different tests - just like this `lighthouse` git repository and this PR has all those separate pipelines for `bbs / ghe / github / gitlab` - those are separate presubmits with separate pipelines.  

So the `.lighthouse/config/lighthouse.yaml` for this git repo could define 6 presubmits there along with (say) 1 templated PipelineRun YAML to implement all the CI/CD for this git repository all in this repository - all in vanilla tekton YAML- rather than in some totally separate Jenkins X git repository using jenkins x pipeline stuff

jstrachan:
maybe another way to think of this PR is - look at how github actions work; folks put into git all the configuration for all the github actions. There's no central magic place in git somewhere that influences how CI/CD works for a repository - its all self contained. This plugin introduces the same idea - for folks using lighthouse with tekton pipelines

jstrachan:
BTW by opt in - I meant - the global configuration of lighthouse + prow defines what plugins are associated with what repositories. So by default noone will use this new plugin. However folks can then configure their global lighthouse configuration to pick which repositories use the `local-config-trigger` plugin - which then means that the presubmit/postsubmit + other configurations are all defined inside the actual git repository (rather than maintained centrally in a huge monster ConfigMap)

eddycharly:
Also, Prow also offers opt-in but not in the form of a plugin, and honestly i dunno if a plugin for that is the best approach.

I mean the logic should be the same in both case, that's just sometimes jobs come from the central repository and sometimes they come from both the central repository and the repo itself (at this point, job config might need to be merged, validated, ensure it's consistent).

Except for the differences in job sources, i don't really see the reasons why the rest of the code should be changed.

eddycharly:
Have you read the prow docs for in repo config ? https://github.com/kubernetes/test-infra/blob/master/prow/inrepoconfig.md

jstrachan:
@eddycharly gotcha. TBH when I created the repo config code I was thinking more broadly about the problem of how to separate the configuration of all of lighthouse into a more decentralised approach - but hadn't started to address integrating that code cleanly into all of lighthouse.

This PR only really tackles the trigger plugin really + handling presubmit/postsubmits nicely - it doesn't try to address keeper or branch protection or anything yet.

To truly make the configuration of all of lighthouse decentralised so we can store the configuration inside each git repository we'll ned a bit more refactoring of the core of lighthouse itself (e.g. to allow post processing the config + plugins structs for each git repository before passing them into the plugins). Then being able to locally configure keeper/branch protection in a local repo config makes more sense.

However from a priorities perspective - being able to define presubmits/postsubmits locally inside a git repository is the most important decentralisation for me; its _very_ common for repositories to want to define its own custom presubmit contexts + pipelines (e.g. this repo). It's much less common to need massively different keeper / branch protection configurations & for them to change much - whereas changing pipelines/contexts is pretty frequent which tends itself to decentralising. After years of working with prow/lighthouse I can't remember configuring one repository differently for keeper for example - yet its very common to want to define different presubmits. e.g. to modify a presubmit pipeline on a PR, just modify the git repository on the PR and you get very fast feedback. Doing it the centralised way is way harder

eddycharly:
But i agree, i can't stand the huge `ConfigMap`s anymore and i am taking a similar approach myself with a `RepoConfig` CRD =)
But that's another story.

jstrachan:
@eddycharly totally agreed with all of that. This PR is the smallest change to start the move for lighthouse to a decentralised approach - tackling the most important things first (presubmit/postsubmits). 

Eventually we could get all configuration decentralised and globally configure what is/isn't centralised and how global + local changes merge together. If we get that far we could make the `local-config-trigger` plugin a noop/alias for `trigger`? But I figured doing all those changes could take a fair amount of time so this is a quick win via an optional plugin while we figure out the full decentralised configuration issues? The plugin is actually pretty tiny so its not a biggie if long term it gets removed once all the other decentralised configuration changes are working?

eddycharly:
I agree 100% for pre/post submit jobs. Not only on the technical side, it's absolutely necessary to enable ownership.

So i don't discuss the need for it at all, i discuss the implementation, IMO only pre/post submits should be configurable in the repo, at least until we find an argument for letting the user configure other things like keeper.

So i guess you could remove all the things that are not related to pre/post submits ?

jstrachan:
I was hoping to get the decentralised code working incrementally too btw; so would be nice to keep that code that can merge the local config into the global config (as we'll need it sometime)

I wonder whats a good middle ground. e.g. maybe the `local-config-trigger` plugin could verify the local configuration and fail if there's anything other than `presubmit/postsubmits` inside? Then folks can accidentally configure - say - keeper? 

Or maybe introduce 2 kinds; `RepositoryConfig` which we don't use yet and `RepositorySubmits` (not sure the name ;) - which is only presubmit/postsubmits? eventually folks could just use `RepositoryConfig` - but for now the `local-config-trigger` plugin would only submit` RepositorySubmits` files?

eddycharly:
Enabling the `local-config-trigger` is going to happen on the centralized side, so it's not better than what prow did https://github.com/kubernetes/test-infra/blob/master/prow/inrepoconfig.md.

Their implementation is simpler too, i still don't see the point in having a separate plugin. In the end trigger doesn't care where the jobs come from.

jstrachan:
ok - it wouldn't be that hard to refactor the webhooks code to allow inrepoconfig I guess?

jstrachan:
yeah - the config/plugins are passed from the webhooks code to the plugins to do stuff via the agent in code like this line:
https://github.com/jenkins-x/lighthouse/blob/9e23a27d8a266db92271a4b9f395936df60bacb3/pkg/webhook/events.go#L157

so we could just have a simple strategy method for creating the agent which if a repository is enabled for `inRepoConfig` we use the repository specify config/plugins structs
https://github.com/jenkins-x/lighthouse/blob/9e23a27d8a266db92271a4b9f395936df60bacb3/pkg/webhook/events.go#L157

Then we'll just need the `inRepoConfig` flag globally + then no need for a custom plugin

hferentschik:
> Have you read the prow docs for in repo config ? https://github.com/kubernetes/test-infra/blob/master/prow/inrepoconfig.md

Would it not make more sense to port the prow inrepoconfig?

abayer:
I'm wondering if it makes sense to _start_ with the Prow repo config approach as the basis for a Lighthouse approach - not just porting it over unchanged, but using that as the starting point?

jstrachan:
@hferentschik we should support something like `inrepoconfig` in the current LH configuration yeah.

The format of the config file for lighthouse inside the repository is less of a concern (e.g. jx / tekton support will differ). The current [RepositoryConfig](https://github.com/jstrachan/lighthouse/blob/local-config-plugin/pkg/plugins/local-config-trigger/test_data/myorg/myrepo/.lighthouse/jenkins-x/lighthouse.yaml) in LH explicitly uses a k8s style layout + versioning so its gonna be easier to work with other k8s technologies and feel similar e.g. tekton / skaffold et al). Plus there's an increasing number of k8s tools (e.g. kpt / Anthos Config Management) that expect `kind` in yaml files) - so using a more k8s native config file format inside the git repositories is a nice idea.

jstrachan:
having backwards compatibility with prow central configuration is useful. However I'm not sure how many folks are fans of the prow configuration file format; I'd personally prefer a nicer more tekton/skaffold style versioned YAML notation using k8s YAML style syntax

abayer:
Oh, I don't care _at all_ about backwards compat with Prow's in-repo config. Preserving backwards compat support for the central config is critical since otherwise we have some hoops to jump through for Jenkins X to keep working, but anything that jx doesn't already care about doesn't have to be Prow-compatible, IMO.

jstrachan:
@abayer totally agreed. So how about we use prow compatible `inrepoconfig` ASAP - then we can keep the current local repository config? 

I might have some time to try a PR tomorrow unless anyone else fancies it?

eddycharly:
I think we should try to be consistent. For example i'm not sure we need different types for in repo config than central config.

Why not stick with the `Presubmit` and `Postsubmit` structs ?

jstrachan:
we're using the same structs for global configuration. its more a question of what the YAML should look like for in repo configurations

eddycharly:
@jstrachan i find it suspicious that you don't have to change anything in keeper too (does keeper needs to know. about the jobs coming from the repo ?).

I am not completely sure but that should be worth double checking. Maybe @abayer can confirm if keeper has to be adapted or not.

jstrachan:
OK I've just pushed a different approach of just adding in-repo configuration. Its using the InRepoConfig from Prow so that the global configuration is prow compatible. There's now no need for a new plugin at all + we just optionally use custom in-repo configuration versions of the config + plugins that are passed into the plugins. 

It turned out to be a pretty simple change to webhooks/event.go - with hindsight I probably should have tried this approach first rather than the plugin ;) 

Here's a little test which verifies that a mostly empty global config (apart from the `InRepoConfig`) manages to find presubmits/postsubmits and triggers etc
https://github.com/jstrachan/lighthouse/blob/local-config-plugin/pkg/webhook/webhook_test.go#L291

eddycharly:
It looks quite better indeed 👍 

When loading the jobs from the repository, it would make sense to take all refs into account before loading. The related code in prow lives [here](https://github.com/kubernetes/test-infra/blob/2efbce017d1a6413667320cd8f979fd40b8ac7b5/prow/config/inrepoconfig.go#L51-L121)

eddycharly:
About the lighthouse and plugins config in the repo, I’m not sure how this would work. All the config could then be different from one branch to another ?

If so, I could just change the config in my branch to get a pr merged without any job, review or approval ?

If the implementation allows such use cases I find it quite dangerous. 

About the pre and post submit jobs, I would definitely merge the refs before loading the config from the repo. If we don’t do that, it opens doors for a pr being green and getting red once it’s merged.

jstrachan:
> About the lighthouse and plugins config in the repo, I’m not sure how this would work. All the config could then be different from one branch to another ?

Maybe we just park the idea of loading the lighthouse and plugins config in repo for now? The main thing I really want is in repo triggers (presubmit/postsubmits)


> About the pre and post submit jobs, I would definitely merge the refs before loading the config from the repo. If we don’t do that, it opens doors for a pr being green and getting red once it’s merged.

Currently we load the pre/post submits from the main branch only so we don't use the contents of the branch or need to merge refs so folks can't disable presubmits on a PR or whatever.

jstrachan:
BTW I've removed everything from the [triggerconfig.ConfigSpec](https://github.com/jstrachan/lighthouse/blob/local-config-plugin/pkg/triggerconfig/types.go#L20) now other than presubmits/postsubmits to keep things nice and simple

jstrachan:
@eddycharly OK I've removed the in repository configuration of lighthouse + plugins. I'm not sure how useful it was anyway & its hard to reason about. 

So things are much simpler now: in repository trigger configuration (presubmits/postsubmits) which are really useful

eddycharly:
Thanks, I will take a look tomorrow.

I feel like we have two points to address in order to make the feature consistent:
- fetching in repo jobs from the main branch only looks not ideal, the in repo jobs should be bound to the branch/pr they belong to
- keeper should know about in repo jobs too, so maybe fetching the in repo config jobs should be built in the config itself and not specific to an agent ?

For the in repo jobs per branch, imagine I have repo with a nodejs app that uses nodejs 8, and I want to migrate to nodejs 12. I should be able to create a branch and use nodejs 12 in my build pipeline for this branch.

pkg/triggerconfig/config_types.go
```diff
@@ -54,18 +53,26 @@ type JobBase struct {
 	// If this field is unspecified or false, a new pod will be created to replace
 	// the evicted one.
 	ErrorOnEviction bool `json:"errorOnEviction,omitempty"`
-	// SourcePath contains the path where this job is defined
-	SourcePath string `json:"-"`
+	// SourcePath contains the path where the PipelineRun YAML is defined
+	SourcePath string `json:"sourcePath,omitempty"`
```
eddycharly, line 57: What if a user specifies both `pipelineRunSpec` and `sourcePath` ?

pkg/triggerconfig/config_types.go
```diff
@@ -54,18 +53,26 @@ type JobBase struct {
...
 	PipelineRunSpec *tektonv1beta1.PipelineRunSpec `json:"pipelineRunSpec,omitempty"`
 	// PipelineRunParams are the params used by the pipeline run
-	PipelineRunParams []job.PipelineRunParam `json:"pipeline_run_params,omitempty"`
+	PipelineRunParams []PipelineRunParam `json:"pipelineRunParams,omitempty"`
 
 	UtilityConfig
 }
 
+// PipelineRunParam represents a param used by the pipeline run
+type PipelineRunParam struct {
```
eddycharly, line 69: Why do we need to duplicate the struct here ?

pkg/triggerconfig/merge/merge.go
```diff
@@ -0,0 +1,46 @@
...
+
+		var ps []job.Postsubmit
+		for _, p := range repoConfig.Spec.Postsubmits {
+			ps = append(ps, ToPostsubmit(p))
+		}
+		cfg.Postsubmits[repoKey] = ps
+	}
+
+	// lets make sure we've got a trigger added
+	idx := len(pluginsCfg.Triggers) - 1
```
eddycharly, line 37: Do we really want to mutate this ?

pkg/triggerconfig/merge/merge.go
```diff
@@ -0,0 +1,46 @@
...
+func ConfigMerge(cfg *config.Config, pluginsCfg *plugins.Configuration, repoConfig *triggerconfig.Config, repoOwner string, repoName string) error {
+	repoKey := repoOwner + "/" + repoName
+	if len(repoConfig.Spec.Presubmits) > 0 {
+		if cfg.Presubmits == nil {
+			cfg.Presubmits = map[string][]job.Presubmit{}
+		}
+
+		var ps []job.Presubmit
+		for _, p := range repoConfig.Spec.Presubmits {
+			ps = append(ps, ToPresubmit(p))
```
eddycharly, line 20: There are probably other constraints to check, like no duplicate job name, etc... recent refactoring should expose a set of methods that can help with individual jobs (`SetDefaults`, `SetRegexes`, `Validate`, etc...) but they also need to be validated as a whole.

abayer:
Yeah, Keeper _has_ to know the presubmit jobs and whether they're required so it can determine whether the required jobs have passed and trigger those jobs if the required jobs have passed but against an older version of the target branch.

pkg/triggerconfig/config_types.go
jstrachan: to use tekton style YAML notation rather than prow style YAML (so the trigger file + tekton pipelines look and feel like they are designed to work together & part of the k8s ecosystem)

pkg/triggerconfig/merge/merge.go
jstrachan: we want to add a trigger for each presubmit/postsubmit yet. Do you have a better suggestion?

pkg/triggerconfig/config_types.go
eddycharly: yeah but why not reuse the existing structs ? is this necessary to duplicate ?

pkg/triggerconfig/merge/merge.go
eddycharly: ah, it's not related to the trigger plugin configuration, i was probably mistaken. sorry.

jstrachan:
OK I've pushed a rebased/merged update which also includes the new `Validate()` methods on config/plugins and I've moved the in repo calculate code into a separate package so its now shared between webhooks + keeper

> * fetching in repo jobs from the main branch only looks not ideal, the in repo jobs should be bound to the branch/pr they belong to

So long as the in-repo presubmits/postsubmits are merged with whatever is in the shared repo, you're cool with loading the in repo triggers from the branch being triggered right? I wasn't sure if you wanted to avoid the possibility of folks removing in-repo presubmit triggers from a PR. 


> * keeper should know about in repo jobs too, so maybe fetching the in repo config jobs should be built in the config itself and not specific to an agent ?

keeper now shares the same in-repo config code as webhooks/plugins/trigger etc.

> For the in repo jobs per branch, imagine I have repo with a nodejs app that uses nodejs 8, and I want to migrate to nodejs 12. I should be able to create a branch and use nodejs 12 in my build pipeline for this branch.

pkg/triggerconfig/merge/merge.go
jstrachan: yeah - its just ensuring that a trigger exists for any presubmit/postsubmit defined in repo basically - its not configuring the trigger plugin

pkg/triggerconfig/config_types.go
jstrachan: I was trying to avoid the `always_run` style of YAML marshalling that the shared prow config uses to use k8s camelCase naming conventions like k8s / tekton / istio etc use such as `alwaysRun`. It adds complexity and confusion though; so I've just removed them all - we're now just using `job.Presubmit` / `job.Postsubmit`directly

Turns out presubmit / postsubmits are fairly simple and there's not many multiple-word fields so its probably not that big of a UX fail really. Longer term we could try move to shorter, simpler property names in Presubmit/Postsubmits maybe anyway ;) (while having a parser to handle the older names). e.g. `always_run` -> `always` and `rerun_command` -> `rerun` maybe?

pkg/triggerconfig/merge/merge.go
jstrachan: we call the `Validate()` methods on lighthouse + prow configs

jstrachan:
@eddycharly OK I just pushed a change where we use the Pull Request ref/sha to query in-repo triggers (presubmit/postsubmits)

eddycharly:
@jstrachan I looked quickly and it’s great that you reused the existing structs.

I still think that we should merge before loading in repo jobs. IMO prow had the correct approach, embedding the feature in the config domain.

jstrachan:
@eddycharly so rather than reading the files from the git branch directly, you'd rather do a git clone + then merge the PR with whatever master is at then read that right?

I guess we'd only need to do that when handling PRs right?

eddycharly:
Yes, the file that we load should be the result of a merge when multiple refs are involved.

When handling a push there will be a single ref so merging will not happen.

eddycharly:
For example you have a job A on master that is not present in a branch.

When the PR is tested, without merging, it doesn’t see job A. Unfortunately the PR breaks job A and when you merge, job A suddenly fails.

Merging before loading the job prevents that, because testing the PR now knows about job A and the failure is detected before the PR gets merged.

jstrachan:
@eddycharly from an implementation perspective, could we load the in repo triggers from the master branch and apply any additions/overrides from the PR/sha? Then a PR doesn't miss anything from the main branch?

jstrachan:
@eddycharly just pushed a simple implementation of loading from the main branch + merging in any changes from the PR/branch

jstrachan:
once this https://github.com/jenkins-x/go-scm/pull/173 merges I've more tests for the PR / main branch merge logic

jstrachan:
@eddycharly here's a little unit test that verifies we merge the triggers from master and the PR together (so that no trigger from master is removed; though it can be changed)
https://github.com/jstrachan/lighthouse/blob/local-config-plugin/pkg/triggerconfig/inrepo/calculate_test.go#L75-L108

also added a validation test that there's no duplicate triggers (presubmit/postsubmits) in a branch when loading - with nice error messages of the problem files + triggers
https://github.com/jstrachan/lighthouse/blob/local-config-plugin/pkg/triggerconfig/scmload/load_triggers_test.go#L32-L42

eddycharly:
> @eddycharly here's a little unit test that verifies we merge the triggers from master and the PR branch/ref together (so that no trigger from master is removed; though it can be changed in a branch/PR)

I would argue that it's not an issue if the job was changed, the PR still has to be reviewed/approved so if the change is not legitimate it should simply not be approved.

eddycharly:
@jstrachan thanks for those improvements.

I think we can start with something like this and eventually improve if necessary.

In the end, what do you think about the prow approach ? Wouldn't it have been easier to just change the `GetPreesubmits` and `GetPostsubmits` methods in the `Config` ?

jstrachan:
> @jstrachan thanks for those improvements.
> 
> I think we can start with something like this and eventually improve if necessary.

great stuff, thanks for all your feedback. We're definitely in a massively better place than when I first raised this PR :) 


> In the end, what do you think about the prow approach ? Wouldn't it have been easier to just change the `GetPreesubmits` and `GetPostsubmits` methods in the `Config` ?

Those methods just filter the presubmits/postsubmits by repo name based on the in memory struct right? There still needs to be code to load the in repo configs from the specific repo + make a new struct - unless I'm misunderstanding which bit of prow code you are talking about? We also need triggers for all presubmit/postsubmits as well; its not just those 2 slices.

I think the LH & prow code for in repo config are pretty similar now really; both using the same `InRepoConfig / Presubmit / PostSubmit` structs etc. The main difference is how the actual repository + ref specific triggers are loaded. Using go-scm to access the config means we don't have to do a full git clone as there's usually only one or two small trigger files to load from each ref/repo.

eddycharly:
> Those methods just filter the presubmits/postsubmits by repo name based on the in memory struct right?

Nope, they checkout and merge refs together then load/validate the in repo config in the centralized config.

`GetPresubmits` calls `getProwYAML`:
```go
func (c *Config) GetPresubmits(gc git.ClientFactory, identifier string, baseSHAGetter RefGetter, headSHAGetters ...RefGetter) ([]Presubmit, error) {
	prowYAML, err := c.getProwYAML(gc, identifier, baseSHAGetter, headSHAGetters...)
	if err != nil {
		return nil, err
	}

	return append(c.PresubmitsStatic[identifier], prowYAML.Presubmits...), nil
}
```

and `getProwYAML` uses a `ProwYAMLGetter` that is responsible of loading the jobs from the repo:
```go
func (c *Config) getProwYAML(gc git.ClientFactory, identifier string, baseSHAGetter RefGetter, headSHAGetters ...RefGetter) (*ProwYAML, error) {
	if identifier == "" {
		return nil, errors.New("no identifier for repo given")
	}
	if !c.InRepoConfigEnabled(identifier) {
		return &ProwYAML{}, nil
	}

	baseSHA, err := baseSHAGetter()
	if err != nil {
		return nil, fmt.Errorf("failed to get baseSHA: %v", err)
	}
	var headSHAs []string
	for _, headSHAGetter := range headSHAGetters {
		headSHA, err := headSHAGetter()
		if err != nil {
			return nil, fmt.Errorf("failed to get headRef: %v", err)
		}
		headSHAs = append(headSHAs, headSHA)
	}
	prowYAML, err := c.ProwYAMLGetter(c, gc, identifier, baseSHA, headSHAs...)
	if err != nil {
		return nil, err
	}

	return prowYAML, nil
}
```

The default `ProwYAMLGetter` is [here](https://github.com/kubernetes/test-infra/blob/ea4b110151f5357d3c87afa7d8ea4208b0e759fc/prow/config/inrepoconfig.go#L51-L121), it does all the checkout + merging refs + loading jobs:
```go
func defaultProwYAMLGetter(
	c *Config,
	gc git.ClientFactory,
	identifier string,
	baseSHA string,
	headSHAs ...string) (*ProwYAML, error) {

	log := logrus.WithField("repo", identifier)
	log.Debugf("Attempting to get %q.", inRepoConfigFileName)

	if gc == nil {
		log.Error("defaultProwYAMLGetter was called with a nil git client")
		return nil, errors.New("gitClient is nil")
	}

	orgRepo := *NewOrgRepo(identifier)
	if orgRepo.Repo == "" {
		return nil, fmt.Errorf("didn't get two results when splitting repo identifier %q", identifier)
	}
	repo, err := gc.ClientFor(orgRepo.Org, orgRepo.Repo)
	if err != nil {
		return nil, fmt.Errorf("failed to clone repo for %q: %v", identifier, err)
	}
	defer func() {
		if err := repo.Clean(); err != nil {
			log.WithError(err).Error("Failed to clean up repo.")
		}
	}()

	if err := repo.Config("user.name", "prow"); err != nil {
		return nil, err
	}
	if err := repo.Config("user.email", "prow@localhost"); err != nil {
		return nil, err
	}
	if err := repo.Config("commit.gpgsign", "false"); err != nil {
		return nil, err
	}

	mergeMethod := c.Tide.MergeMethod(orgRepo)
	log.Debugf("Using merge strategy %q.", mergeMethod)
	if err := repo.MergeAndCheckout(baseSHA, string(mergeMethod), headSHAs...); err != nil {
		return nil, fmt.Errorf("failed to merge: %v", err)
	}

	prowYAMLFilePath := path.Join(repo.Directory(), inRepoConfigFileName)
	if _, err := os.Stat(prowYAMLFilePath); err != nil {
		if os.IsNotExist(err) {
			log.Debugf("File %q does not exist.", inRepoConfigFileName)
			return &ProwYAML{}, nil
		}
		return nil, fmt.Errorf("failed to check if file %q exists: %v", inRepoConfigFileName, err)
	}

	bytes, err := ioutil.ReadFile(prowYAMLFilePath)
	if err != nil {
		return nil, fmt.Errorf("failed to read %q: %v", inRepoConfigFileName, err)
	}

	prowYAML := &ProwYAML{}
	if err := yaml.Unmarshal(bytes, prowYAML); err != nil {
		return nil, fmt.Errorf("failed to unmarshal %q: %v", inRepoConfigFileName, err)
	}

	if err := DefaultAndValidateProwYAML(c, prowYAML, identifier); err != nil {
		return nil, err
	}

	log.Debugf("Successfully got %d presubmits and %d postsubmits from %q.", len(prowYAML.Presubmits), len(prowYAML.Postsubmits), inRepoConfigFileName)
	return prowYAML, nil
}
```

jstrachan:
> > The main difference is how the actual repository + ref specific triggers are loaded. Using go-scm to access the config means we don't have to do a full git clone as there's usually only one or two small trigger files to load from each ref/repo.
> 
> This would be easily feasible with the prow approach, it's really just another `ProwYAMLGetter` implementation that calls the provider API instead of git clone/git merge.

the thing I like about the current LH codebase is simplicity; for each webhook on each repo there's a single call to create theLH config / prow configs from the repository/sha without worrying about caching or anything; the structs are then shared across all invocations of all plugins - irrespective of what methods the plugins call. 

i.e. as soon as the webhook invokes plugins for a repo/sha they all share the same in memory structs. My worry with adding the loading code inside those 2 method calls inside the config object and then exposing that API to all plugins whenever some code wants a presubmit/postsubmit is having lots of git clone / go-scm calls for the same webhook. It also isolates the code so we can make any change we want to anything in the combined lighthouse + prow config based on in-repo data in a single place (e.g. enabling triggers if they are missing on a repo). Over time we may think of other things we want to change

eddycharly:
@jstrachan you need to rebase

jstrachan:
@eddycharly done

abayer:
@jstrachan Any chance you could squash down the commits?

jstrachan:
@abayer done

eddycharly:
/lgtm
/hold

LGTM from me but i'll leave final approval to @abayer

pkg/config/lighthouse/config.go
```diff
@@ -105,3 +108,55 @@ func (c *Config) Parse() error {
...
+// InRepoConfig to enable configuration inside the source code of a repository
+type InRepoConfig struct {
+	// Enabled describes whether InRepoConfig is enabled for a given repository. This can
+	// be set globally, per org or per repo using '*', 'org' or 'org/repo' as key. The
+	// narrowest match always takes precedence.
+	Enabled map[string]*bool `json:"enabled,omitempty"`
+	// AllowedClusters is a list of allowed clusternames that can be used for jobs on
+	// a given repo. All clusters that are allowed for the specific repo, its org or
+	// globally can be used.
+	AllowedClusters map[string][]string `json:"allowed_clusters,omitempty"`
```
hferentschik, line 121: @jstrachan, what is this about? How is the ability to configure the Lighthouse config in the repository tied to a specific cluster?

jstrachan: that came from the Prow struct. We don’t use it in any way right now

jstrachan: happy to delete it - but there seemed interest in the discussions on this PR of using Prow’s `InRepoConfig` so that the shared config is similar is to prows

pkg/keeper/keeper.go
```diff
@@ -73,6 +74,7 @@ type scmProviderClient interface {
 	GetRepositoryByFullName(string) (*scm.Repository, error)
 	ListAllPullRequestsForFullNameRepo(string, scm.PullRequestListOptions) ([]*scm.PullRequest, error)
 	CreateComment(owner, repo string, number int, isPR bool, comment string) error
+	ToScmClient() *scm.Client
```
hferentschik, line 77: This seems odd to have in the interface, or not?

pkg/config/lighthouse/config.go
hferentschik: I see.

hferentschik: hmm, similar yes, but if something is not used my initial gut feeling would say not to add it (easier to add something later when we need it opposed to removing something). @eddycharly - wdyt?

jstrachan: removed for now. We can easily git revert if we choose to bring it back

pkg/keeper/keeper.go
jstrachan: OK switched to the _scmProviderClient_ interface providing the API instead of go-scm's `Client` interface

abayer:
Failures are due to athens proxy issues - https://github.com/jenkins-x/lighthouse/pull/1064/files#diff-37aff102a57d3d7b797f152915a6dc16 will fix them.

jstrachan:
@hferentschik @abayer anything else needed on this PR?

abayer:
Nah, I think we're good!

/hold cancel
/lgtm

PR #1005, opened 2020-09-03
chore: make webhook event handlers private

eddycharly:
This PR makes webhook event handlers private (they don't need to be public)


abayer:
/lgtm

I wonder why they were public in the first place...

eddycharly:
Yeah, it looked strange to have these functions public, it's an implementation detail by nature.

eddycharly:
It would be nice to remove the legacy_helpers too. It itches me =)

PR #1007, opened 2020-09-03
feat: Tests/official support for Gitea

abayer:
Yay! Gitea 1.13 finally released, so this is ready to go. =)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,5 +1,4 @@
-# Make
- does not offer a recursive wildcard function, so here's one:
+# Make does not offer a recursive wildcard function, so here's one:
 rwildcard=$(wildcard $1$2) $(foreach d,$(wildcard $1*),$(call rwildcard,$d/,$2))
 
 PROJECT := github.com/jenkins-x/lighthouse
--- a/bdd/gitea/ci.sh
+++ b/bdd/gitea/ci.sh
@@ -159,7 +159,7 @@ if kubectl get configmap config ; then
 fi
 
 # Run the test - we probably want something here to capture controller logs but we'll deal with that in a bit.
-make run-e2e-tests
+make run-tekton-e2e-tests
 
 bdd_result=$?
 if [[ $bdd_result != 0 ]]; then
--- a/bdd/gitea/ci.sh
+++ b/bdd/gitea/ci.sh
@@ -1,0 +1,170 @@
+#!/usr/bin/env bash
+set -e
+
+export E2E_PRIMARY_SCM_USER=lighthouse-bot
+export E2E_APPROVER_SCM_USER=approver
+
+# Error out if required environment variables aren't set.
+missingEnvVarMessages=()
+if [ -z "${VERSION}" ]; then
+  missingEnvVarMessages+=( "VERSION for Lighthouse images not set" )
+fi
+if [ -z "${E2E_GIT_KIND}" ]; then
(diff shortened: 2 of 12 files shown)


abayer:
I'm using the `tekton` context for the moment.

abayer:
Blech, there seems to be a problem with getting the token set properly in the Gitea client under the hood. Working on it.

abayer:
So current situation is that it's not populating the `OWNERS` link from `Needs approval from an approver in each of these files` in the initial comment, and that means I have to wade into the hell of owners file parsing and such. yay. =)

abayer:
OK! So the problem is that we have to have a hack to do the `PullRequest.ListChanges` function with Gitea, because they don't have an equivalent endpoint. So what I decided to do was get the `.patch` and parse that into the changes we need. Hoorah! Should work! But https://github.com/go-gitea/gitea/issues/10923 - can't get the `.diff` or `.patch` with an api token, just via basic authentication. Except! https://github.com/go-gitea/gitea/pull/11751 - it's fixed, now there are such endpoints...but they aren't released yet. We need to wait for 1.13.0. Baaaaah.

abayer:
I hacked around it by using the latest pre-release image of gitea, but we aren't going to want to merge this until it's actually released. I also opened https://gitea.com/gitea/go-sdk/pulls/398 so we don't have to do hacks for `ListChanges`. Next up is figuring out why Keeper refuses to start...

abayer:
yaaaaaay everything passed.

abayer:
So this works - it's just waiting on Gitea 1.13.0 releasing for the `.patch` endpoint (so we can calculate the list of changed files in a PR). I'll keep it on WIP until then.

eddycharly:
DeleteRef was not implemented for Gitea, the branchcleaner plugin does not work. I opened a PR (https://github.com/jenkins-x/go-scm/pull/177)

eddycharly:
@abayer any idea how to take care of the branch name when deleting a branch in the branchcleaner plugin ?
https://github.com/jenkins-x/lighthouse/blob/5fd1d1f5e64ec136ed18591181f1e863e0667cca/pkg/plugins/branchcleaner/branchcleaner.go#L63

Adding the `heads/` prefix does not work for gitea.

abayer:
Hrm, I'll take a look. I'm actually kinda regretting using the Gitea Go SDK behind the scenes now because it doesn't work the way the other providers do re: pagination, so I may just rewrite everything today. =)

eddycharly:
Yeah, in kloops i introduced an `ScmTools` interface to accommodate for these differences (https://github.com/eddycharly/kloops/blob/520d6dc1e92a1be8979e082c47ba908ebb31623a/pkg/scmprovider/tools.go#L22-L27).

abayer:
Let's see how far I get with rewriting - if I don't get it done before Gitea 1.13 is released, we can special case like you're doing in kloops.

abayer:
Ah-ha, the new version of gitea/go-sdk has `http.Response` (well, a wrapper around it, but...) returns now, so tada, https://github.com/jenkins-x/go-scm/pull/180

hferentschik:
@abayer what's the state on this one? I guess at the very least is seems to require a rebase.

abayer:
@hferentschik Waiting for gitea to release 1.13 for https://github.com/go-gitea/gitea/pull/11751, which we’re currently using to construct the list of changed files in a PR.

abayer:
Well, blech, the Gitea chart is currently broken on k8s 1.18 - https://gitea.com/gitea/helm-chart/issues/77.

PR #1024, opened 2020-09-09
refactor: manage commands in Plugin descriptor

eddycharly:
This PR changes the way plugin commands are managed by the plugin engine.

Before, the commands were registered separately from the help provider, it was simply a `GenericCommentEventHandler` and there was no guarantee that the command and the help provider were consistent.

Now, the commands are part of the definition of a plugin, the help is moved in the command definition, this makes it easier to keep the command and its help consistent.
The command registration can specify the regex to use and a filter func, these parameters will be taken care of by the plugin engine automatically and the regex match will be passed to the handler when being invoked by the plugin engine.

If the command definition does not specify a regex or a filter func, it will behave like before (receive all events and the filtering/matching has to be done by the handler).

### DIFF
--- a/pkg/plugins/approve/approve.go
+++ b/pkg/plugins/approve/approve.go
@@ -94,20 +94,33 @@ type state struct {
 	htmlURL   string
 }
 
-func init() {
-	plugins.RegisterPlugin(
-		PluginName,
-		plugins.Plugin{
-			Description: `The approve plugin implements a pull request approval process that manages the '` + labels.Approved + `' label and an approval notification comment. Approval is achieved when the set of users that have approved the PR is capable of approving every file changed by the PR. A user is able to approve a file if their username or an alias they belong to is listed in the 'approvers' section of an OWNERS file in the directory of the file or higher in the directory tree.
+var (
+	plugin = plugins.Plugin{
+		Description: `The approve plugin implements a pull request approval process that manages the '` + labels.Approved + `' label and an approval notification comment. Approval is achieved when the set of users that have approved the PR is capable of approving every file changed by the PR. A user is able to approve a file if their username or an alias they belong to is listed in the 'approvers' section of an OWNERS file in the directory of the file or higher in the directory tree.
 <br>
 <br>Per-repo configuration may be used to require that PRs link to an associated issue before approval is granted. It may also be used to specify that the PR authors implicitly approve their own PRs.
 <br>For more information see <a href="https://git.github.com/jenkins-x/lighthouse/pkg/prow/plugins/approve/approvers/README.md">here</a>.`,
-			HelpProvider:          helpProvider,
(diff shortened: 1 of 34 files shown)


eddycharly:
/hold

must come after https://github.com/jenkins-x/lighthouse/pull/1023

eddycharly:
Ok, i probably broke something...

abayer:
The Jenkins X contexts never merged the test PR - everything before then worked, so I'm not sure what you changed could have caused that...

eddycharly:
Ok, it's probably around the lgtm or approve plugins then...

eddycharly:
/test tekton

this one was passing...

abayer:
lgtm and approve are working - the last thing that happens is that it `/override`s, which it did successfully, and then waits for Keeper to merge the PR. That's what's weird - but I'll dig into logs if it continues.

abayer:
Tekton failed on `/uncc` - but since everything else has passed that cleanly before, I think it may have just been random noise.

abayer:
Ah-ha, it's `/override` - it briefly sets the status correctly but then it gets overwritten with a failure and so things can't merge.

abayer:
override has a history of weird behavior like this, for what it's worth.

abayer:
aaaaaah `/override` is _also_ triggering a run, not just creating a `LighthouseJob` that just sets a status and doesn't do anything else.

abayer:
Actually, it might not be `override`, but there are 7 runs being created for the PR, and there should be 3, so it looks like we're spawning more runs than there should be, and while one of the three that seemed to be running at the same time failed and the test saw that and proceeded to override, another of the remaining two runs failed after the override triggered and confused things.

abayer:
Verified - it looks like `/test this` and `/retest` each triggered 3 runs.

eddycharly:
Ok, understood, there are 3 commands in the trigger plugin...

eddycharly:
Each command gets the event but there is no logic to acknowledge the event was consumed... gotta change that, the unit tests would have failed but they didn't. I was surprised but didn't pay more attention.

Thanks, gonna fix that.

eddycharly:
It should be fixed this time.

abayer:
Failure is in linting:

```
LINTING
/home/jenkins/go/bin/golint -set_exit_status ./...
pkg/plugins/plugins.go:89:1: exported method Plugin.InvokeCommand should have comment or be unexported
Found 1 lint suggestions; failing.
make: *** [lint] Error 1
```

eddycharly:
yep, forgot those...
hopefully other pipelines will be green 🤞

abayer:
/lgtm

Yup, looks all good now. Woo!

eddycharly:
/hold cancel

I'll try to harden the plugin contract, making the regex mandatory first, then rework the command/command help relations.

PR #1031, opened 2020-09-09
refactor: dog plugin

eddycharly:
This PR refactors the dog plugin.

It adds separate commands for all commands regular expressions `/this-is-fine`, `/this-is-not-fine` and `/this-is-unbearable`.

### DIFF
--- a/pkg/plugins/dog/dog.go
+++ b/pkg/plugins/dog/dog.go
@@ -49,25 +49,71 @@ const (
 	pluginName    = "dog"
 )
 
-var (
-	plugin = plugins.Plugin{
+func createPlugin(p pack) plugins.Plugin {
+	return plugins.Plugin{
 		Description: "The dog plugin adds a dog image to an issue or PR in response to the `/woof` command.",
 		Commands: []plugins.Command{{
-			GenericCommentHandler: handleGenericComment,
-			Filter:                func(e scmprovider.GenericCommentEvent) bool { return e.Action == scm.ActionCreate },
+			Filter: func(e scmprovider.GenericCommentEvent) bool { return e.Action == scm.ActionCreate },
+			Regex:  match,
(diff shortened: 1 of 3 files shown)


pkg/plugins/dog/dog.go
```diff
@@ -49,25 +49,71 @@ const (
...
 		Commands: []plugins.Command{{
-			GenericCommentHandler: handleGenericComment,
-			Filter:                func(e scmprovider.GenericCommentEvent) bool { return e.Action == scm.ActionCreate },
+			Filter: func(e scmprovider.GenericCommentEvent) bool { return e.Action == scm.ActionCreate },
+			Regex:  match,
+			GenericCommentHandler: func(_ []string, pc plugins.Agent, e scmprovider.GenericCommentEvent) error {
+				return handle(pc.SCMProviderClient, pc.Logger, &e, p)
+			},
+			Help: []pluginhelp.Command{{
+				Usage:       "/(lh-)?(woof|bark)",
```
hferentschik, line 62: can you refresh my memory, why do we need a 'lh-*' version for everything again?

eddycharly: Dunno, I wasn’t there when it was added.

I believe there is one plugin that behaves differently when the lh- prefix is set, I guess it was added everywhere for consistency. 

I would vote for removing it when not necessary though.

abayer: Because Gitlab hijacks some slash commands for its own [quick actions](https://docs.gitlab.com/ee/user/project/quick_actions.html), and we never get notified. In practice, we don’t need the lh prefix for everything, just commands that also are quick actions, but I opted to play it safe and add lh prefixes for every command just in case Gitlab eventually adds conflicting quick actions.

eddycharly: Good to know ! Thanks

hferentschik: thanks for the heads-up.

PR #1033, opened 2020-09-10
feat: Port Prow Jenkins client and controller to Lighthouse

### LINKED ISSUE #885: Support Jenkins job execution engine

hferentschik:
fixes #885

This ports the Prow Jenkins controller to Lighthouse, adjusting types, etc.


Makefile
```diff
@@ -64,80 +124,22 @@ $(GOSEC):
...
-.PHONY: gc-jobs
-gc-jobs:
-	$(GO) build -i -ldflags "$(GO_LDFLAGS)" -o bin/$(GCJOBS_EXECUTABLE) $(GCJOBS_MAIN_SRC_FILE)
-
-.PHONY: tekton-controller
-tekton-controller:
-	$(GO) build -i -ldflags "$(GO_LDFLAGS)" -o bin/$(TEKTONCONTROLLER_EXECUTABLE) $(TEKTONCONTROLLER_MAIN_SRC_FILE)
-
-.PHONY: compile-e2e
-compile-e2e:
```
abayer, line 99: We still need this and `run-e2e-tests` (though the latter also will need to be reworked to have separate targets for running the Tekton tests and the eventual Jenkins tests).

hferentschik: Interesting. I thought I search in the whole project for occurrences. Let me bring it back.

abayer:
So I see you're not tackling the `LighthouseJob.Status.Activity` yet - for now, we can probably just paper over that by adding something like this (with the stage's status changing as appropriate):
```
lighthouseJob.Status.Activity = &v1alpha1.ActivityRecord{
    Stages: []*v1alpha1.ActivityStageOrStep{{
        Name: "build",
        Status: v1alpha1.RunningState, // or v1alpha1.PendingState, AbortedState, FailureState, etc
    }
}
```

I'll follow up later with logic using https://github.com/jenkinsci/pipeline-graph-analysis-plugin (which is used by Blue Ocean to get stage/step information) to fully populate the `ActivityRecord`, but something simple for now is fine.

I'm going to be ok merging this without an end to end test, but that should definitely be next up after this.

hferentschik:
> So I see you're not tackling the LighthouseJob.Status.Activity yet - for now, we can probably just paper over that by adding something like this (with the stage's status changing as appropriate):

No ;-), mainly because I did not know I had to. Still getting to know all the bits and pieces here. Now that you mention `Activity` I recall this being a thing in Jenkins X. So is this `ActivityRecord` an addition to Lighthouse which did not come from Prow, right? Was there not enough status progress information as it was or why the `Activity`?

> ```
> lighthouseJob.Status.Activity = &v1alpha1.ActivityRecord{
>     Stages: []*v1alpha1.ActivityStageOrStep{{
>        Name: "build",
>        Status: v1alpha1.RunningState, // or v1alpha1.PendingState, AbortedState, FailureState, etc
>   }
>}
>```

Ok, let me have a look at this. 

> I'll follow up later with logic using https://github.com/jenkinsci/pipeline-graph-analysis-plugin (which is used by Blue Ocean to get stage/step information) to fully populate the ActivityRecord, but something simple for now is fine.

Ahh, I see. There are probably more missing pieces and improvements we can do. This is really a minimal port so far.

> I'm going to be ok merging this without an end to end test, but that should definitely be next up after this.

+1. I can go either way. I think this makes a good start and one can tackle smaller issues one at a time now. That said, I am happy to add some initial test setup. Question is how we go about this? Do you have any ideas? Do you want to standup a Jenkins instance on the fly or can we use an existing one? Do we continue adding a new Jenkins X context for that or try to do it somewhere else?

hferentschik:
@abayer, still fiddling a bit the Helm chart a bit and then there are some other minor tweaks I want to add and I want to add something to the docs/README.  Overall it's getting there though.

When it comes to the Helm chart and the initial install we need to think a bit how we best want to do this. There are several traps you can fall into, especially if you want to back the config via a git repo. I think it would be nice to have an "easy" install just via Helm. Ideally that's how it should be. On the other hand this means that there is a bit of manual and error prone work required by the user when he wants to use Git as config store. Having an installer sucks a bit since it takes away from the simple `helm install`. An alternative could be to bake something into the binary which does it for once triggered. Anyways, thought thinking out loud here.

pkg/config/lighthouse/config.go
```diff
@@ -32,7 +32,7 @@ type Config struct {
 	BranchProtection branchprotection.Config `json:"branch-protection,omitempty"`
 	Orgs             map[string]org.Config   `json:"orgs,omitempty"`
 	// TODO: Move this out of the main config.
-	JenkinsOperators []JenkinsOperator `json:"jenkins_operators,omitempty"`
+	Jenkinses []JenkinsConfig `json:"jenkins_operators,omitempty"`
```
hferentschik, line 35: @abayer, here is another thing I would like to adjust. I tried to get away from using the term `jenkins_operators` wherever I could. I think it is misleading. The only place left is here which unfortunately is visible to the user, since  `jenkins_operators` will be used in the config YAML. 

I initially used `jenkins`, but then noticed that other fields also did not get renamed, eg see Keeper a few lines above which still maps to `tide` in the config.  Was there the idea that the configurations should stay for some reason compatible? I find this highly confusing. We talk about Keeper, but configure "Tide" in the YAML. I think that was ok while Lighthouse was a pure Jenkins X internal thing, but if it is supposed to stand on its own feet, we need to address this somehow. 

@abayer, do you have any ideas/plans around this? Should I keep `jenkins_operators` or can I switch to `jenkins` or `jenkinses`?

abayer:
@hferentschik Yeah, `ActivityRecord` is descended from jx's `PipelineActivity` - currently we're just using it to determine what stages/tasks/etc are currently running for putting in the GitHub/etc commit status description, but we're also sending the `ActivityRecord` to external plugins (if they're so configured) so that they can do things like slack notifications, detailed status reporting, etc.

Re testing - I think spinning up a new Jenkins instance in a new cluster is probably the way to go. We could use [Jenkins CasC](https://www.jenkins.io/projects/jcasc/) to prebake an image with the configuration we need and then install Jenkins via helm.

pkg/config/lighthouse/config.go
abayer: I've generally gone with the idea of keeping any field that Jenkins X cares about the same, but configuration for anything that jx doesn't care about (like Jenkins) can be different from Prow.

pkg/engines/jenkins/controller.go
```diff
@@ -0,0 +1,485 @@
...
+		// Otherwise, abort it.
+		if buildExists {
+			if err := c.jenkinsClient.Abort(getJobName(&toCancel.Spec), &build); err != nil {
+				c.log.WithError(err).WithFields(jobutil.LighthouseJobFields(&toCancel)).Warn("Cannot cancel Jenkins build")
+			}
+		}
+
+		prevState := toCancel.Status.State
+		toCancel.Status.State = v1alpha1.AbortedState
+		lighthouseJob.Status.Activity = &v1alpha1.ActivityRecord{
```
hferentschik, line 274: @abayer Is this what you had in mind?

abayer: Exactly!

abayer:
/retest

Athens proxy barfed. =)

hferentschik:
> Re testing - I think spinning up a new Jenkins instance in a new cluster is probably the way to go. We could use Jenkins CasC to prebake an image with the configuration we need and then install Jenkins via helm.

So lets see, if I get this right:

- Create a new Jenkins X context, aka `jenkins-x-jenkins.yml`
    - difference to the others is that in the end it calls something like `bdd/jenkins/ci.sh`
- Create `bdd/jenkins/ci.sh`
    - create the test cluster
    - install Jenkins 
    - install Lighthouse
- Create jenkins_test.go
    - create a test repo
    - create the Jenkins  test pipeline 
    - configure Lighhouse for use of Jenkins and the test repo
    - make a change to the git repo
    - verify Jenkins jobs runs

I think you implied to pre-bake a Jenkins image. I for my part was thinking to install Jenkins via Helm as well and then use its REST API later to create the Jenkins Job (either in `ci.sh` or as suggested above in `jenkins_test.go`).

@abayer, does this sound right? Shall I try tackling this as part of this PR or better do do this in a follow up issue?

abayer:
Yup, that's basically it. What I've done for testing the context before it's actually ready is to temporarily hijack an existing context. And yeah, it can wait and be done as a followup.

And re prebaking and helm - I was actually thinking a combination, prebaking an image and then using the helm chart to install that image, so that things like creds are already defined and we don't need to jump through hoops to generate them, but that's just an implementation detail. =)

hferentschik:
> Yup, that's basically it. What I've done for testing the context before it's actually ready is to temporarily hijack an existing context. And yeah, it can wait and be done as a followup.

FYI - https://github.com/jenkins-x/lighthouse/issues/1048

pkg/engines/jenkins/controller.go
```diff
@@ -0,0 +1,456 @@
...
+			WithField("to", lighthouseJob.Status.State).Info("Transitioning states.")
+		// make sure to set an activity record this job state update
+		c.addActivity(&lighthouseJob)
+		_, err = c.lighthouseClient.UpdateStatus(&lighthouseJob)
+	}
+
+	return err
+}
+
+func (c *Controller) addActivity(lighthouseJob *v1alpha1.LighthouseJob) {
```
hferentschik, line 436: @abayer - that seems to cut it for now. 

That said, is it not a bit odd that the _ActivityRecord_ just mirrors what's anyways already stored on the LighthouseJob?

pkg/apis/lighthouse/v1alpha1/types.go
```diff
@@ -393,3 +412,10 @@ func (a *ActivityRecord) RunningStages() []string {
 	}
 	return running
 }
+
+// JenkinsSpec is optional parameters for Jenkins jobs.
+// Currently, the only parameter supported is for telling
+// jenkins-operator that the job is generated by the https://go.cloudbees.com/docs/plugins/github-branch-source/#github-branch-source plugin
+type JenkinsSpec struct {
+	GitHubBranchSourceJob bool `json:"github_branch_source_job,omitempty"`
```
hferentschik, line 420: @abayer - that's GitHub specific really. Apparently for PR builds the code relies on the Git Branch Source plugin. In the long run we need something else here (provider independent), the question is whether we should have this interim?

abayer: Rename it to `BranchSourceJob` - the mechanics of Jenkins branch source are the same from our perspective, so let's keep it around.

hferentschik: done

abayer:
dammit, the jsonpatch thing came back!

hferentschik:
> dammit, the jsonpatch thing came back!

doh, how the heck!?

hferentschik:
@abayer, anything else you want me to do for now?

abayer:
@hferentschik Looks like it just needs a rebase!

hferentschik:
> Looks like it just needs a rebase!

Bugger that. Someone came in between ...

hferentschik:
@abayer - done

hferentschik:
I guess I am up for another rebase ...

pkg/engines/jenkins/controller.go
```diff
@@ -0,0 +1,456 @@
...
+		if cfg.LabelSelectorString == c.selector {
+			return cfg.Controller
+		}
+		configured = append(configured, cfg.LabelSelectorString)
+	}
+	if len(c.selector) == 0 {
+		c.log.Panicf("You need to specify a non-empty --label-selector (existing selectors: %v).", configured)
+	} else {
+		c.log.Panicf("No config exists for --label-selector=%s.", c.selector)
+	}
```
LinuxSuRen, line 118: I'm curious about this part. It looks like no matter `c.selector` is what it always panic.

PR #1035, opened 2020-09-10
refactor: stage and pony plugins

eddycharly:
This PR refactors stage and pony plugins


abayer:
/lgtm

As always, thanks for making the code less confusing and weird. =)

PR #1038, opened 2020-09-10
refactor: milestone plugin

eddycharly:
This PR refactors the milestone plugin.

The `Milestone` service doesn't exist in `go-scm` though, not sure if this plugin works ?

### DIFF
--- a/pkg/plugins/milestone/milestone.go
+++ b/pkg/plugins/milestone/milestone.go
@@ -56,9 +56,11 @@ var (
 		Description:        "The milestone plugin allows members of a configurable GitHub team to set the milestone on an issue or pull request.",
 		ConfigHelpProvider: configHelp,
 		Commands: []plugins.Command{{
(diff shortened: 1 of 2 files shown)


hferentschik:
> The Milestone service doesn't exist in go-scm though, not sure if this plugin works ?

If it does not work, I'd rather remove this altogether? @abayer wdyt?

abayer:
I'd lean towards actually implementing milestones (in GitHub and Gitea, epics in GitLab) in go-scm. I might do that today.

hferentschik:
> I'd lean towards actually implementing milestones in go-scm. I might do that today.

so it is a concept available across scm providers? Fair enough.

abayer:
/lgtm

I'll get a PR up later today for the milestone API and related stuff in go-scm, and then a followup to actually implement it here.

eddycharly:
I’m thinking about adding providers exclusion for plugins. It could be handy if the api is not available with every provider.

PR #1051, opened 2020-09-12
refactor: sigmention plugin

eddycharly:
This PR refactors the sigmention plugin.

The sigmention plugin does not follow the command concept, it catches all generic comment events and doesn't require matching a regex with a command like format.

To address this behavior i decoupled the `CommandEventHandler` and `GenericCommentEventHandler`. The former receives a regex match while the latter just receives the event and doesn't have an associated command help.

### DIFF
--- a/pkg/plugins/dog/dog_test.go
+++ b/pkg/plugins/dog/dog_test.go
@@ -230,7 +230,7 @@ func TestHttpResponse(t *testing.T) {
 			Logger:            logrus.WithField("plugin", pluginName),
 		}
 		plugin := createPlugin(realPack(ts.URL))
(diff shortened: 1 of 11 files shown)


abayer:
Legit vet failure in override:
```
vet: pkg/plugins/override/override_test.go:432:47: cannot use (func(handler plugins.GenericCommentHandler, e *scmprovider.GenericCommentEvent, match []string) error literal) (value of type func(handler plugins.GenericCommentHandler, e *scmprovider.GenericCommentEvent, match []string) error) as func(plugins.CommandEventHandler, *scmprovider.GenericCommentEvent, []string) error value in argument to plugin.InvokeCommandHandler
```

eddycharly:
Thanks, it should be fixed now.

PR #1055, opened 2020-09-14
refactor: refactor the plugin command system

eddycharly:
This PR refactors the plugin command system.

It makes the command system non regex dependent. A command has the format `/[PREFIX]COMMAND [ARG]`, this is declared at the command level and is also used to generate the command usage and examples help.
The command description is also added to the command and is used to generate the command help.

I added unit tests for the `Command` struct.

I also clean up some obsolete code that was created at the start of refactoring and not used anymore.

I'll have to write a doc on how to write a plugin command.

### DIFF
--- a/pkg/plugins/lifecycle/reopen_test.go
+++ b/pkg/plugins/lifecycle/reopen_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package lifecycle
 
 import (
+	"fmt"
 	"testing"
 
 	"github.com/jenkins-x/go-scm/scm"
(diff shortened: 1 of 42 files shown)


pkg/plugins/lifecycle/reopen_test.go
```diff
@@ -65,42 +66,42 @@ func TestReopenComment(t *testing.T) {
...
+		// 	body:          "does not matter",
+		// 	commenter:     "random-person",
+		// 	shouldReopen:  false,
+		// 	shouldComment: false,
+		// },
+		// {
+		// 	name:          "re-open by author",
+		// 	action:        scm.ActionCreate,
+		// 	state:         "closed",
+		// 	body:          "/reopen",
```
abayer, line 82: Why are these commented out?

eddycharly: Good catch, I commented them out while chasing a regex match issue.
Fixed.

abayer:
/lgtm

Very nice!

PR #1056, opened 2020-09-14
feat: add scm providers exclusion support to plugins

eddycharly:
This PR adds scm providers exclusion support for plugins.

I also moved the `Plugin` struct to its own file and started adding unit tests.

I also added the `ExcludedProviders` field in the `PluginHelp` struct.


eddycharly:
/hold
I would like to get the check implemented at the invoke level to make testing easier.

eddycharly:
/hold cancel

Let's go like this, i have a few other PRs to finalize and i will address testing in another refactoring pass.

PR #1058, opened 2020-09-14
feat: filter bot comments at the plugin engine level

eddycharly:
This PR adds bot comments filtering at the plugin engine level.

### DIFF
--- a/pkg/plugins/approve/approve.go
+++ b/pkg/plugins/approve/approve.go
@@ -99,6 +99,7 @@ var (
 <br>Per-repo configuration may be used to require that PRs link to an associated issue before approval is granted. It may also be used to specify that the PR authors implicitly approve their own PRs.
 <br>For more information see <a href="https://git.github.com/jenkins-x/lighthouse/pkg/prow/plugins/approve/approvers/README.md">here</a>.`,
 		ConfigHelpProvider: configHelp,
+		ReceiveBotComments: true,
 		ReviewEventHandler: handleReviewEvent,
 		PullRequestHandler: handlePullRequestEvent,
 		Commands: []plugins.Command{{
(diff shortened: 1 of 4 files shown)


eddycharly:
/hold

Must come after https://github.com/jenkins-x/lighthouse/pull/1056

eddycharly:
Something going wrong ?
Probably one plugin needs to receive bot events ?

abayer:
Our tests for verifying that `/lgtm` will result in a message showing up that you can't lgtm your own PR are failing with this, because the hook gets discarded before getting to `lgtm`.

eddycharly:
This is because the test is executed with the bot user ?

abayer:
@eddycharly Yup. We have two users - the bot user and an approver user. Either we'd need a third user or we'd need `/lgtm` to get through when sent by the bot, which is what you just changed, so we should be good. =)

pkg/plugins/approve/approve.go
```diff
@@ -99,6 +99,7 @@ var (
 <br>Per-repo configuration may be used to require that PRs link to an associated issue before approval is granted. It may also be used to specify that the PR authors implicitly approve their own PRs.
 <br>For more information see <a href="https://git.github.com/jenkins-x/lighthouse/pkg/prow/plugins/approve/approvers/README.md">here</a>.`,
 		ConfigHelpProvider: configHelp,
+                ReceiveBotComments: true,
```
abayer, line 102: oops, formatting. =)

abayer:
Ah, but other ones are going to fail too - `/cc`, `/test`, `/retest`, `/override`...probably best to hold off on this for the moment and think about it more.

eddycharly:
Ok, if receiving comments from the bot user doesn’t hurt, I guess it’s not worth filtering then.

eddycharly:
@abayer shall i drop completely this PR ? or is there some use case where we want to filter out bot comments ?

abayer:
@eddycharly Yeah, I'd say let's drop it for now.

eddycharly:
Ok, closing.
/close

PR #1059, opened 2020-09-14
feat: add command condition builder

eddycharly:
This PR adds command condition builder support.

This allows a simpler syntax and reuse of conditions across plugins.

### DIFF
--- a/pkg/plugins/approve/approve.go
+++ b/pkg/plugins/approve/approve.go
@@ -26,15 +26,14 @@ import (
 	"time"
 
 	"github.com/jenkins-x/go-scm/scm"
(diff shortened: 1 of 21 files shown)


eddycharly:
/hold

Must come after #1058

eddycharly:
Next:
- implement error handling at the plugin engine level instead of having individual plugins handle error separately (maybe create comment with the error message or something like this)
- rework testing
- docs

eddycharly:
Probably failing because of #1058 :(

eddycharly:
/hold cancel

I removed the bot comment filtering part

PR #1072, opened 2020-09-21
 fix: merging postsubmits & presubmits together by name to avoid duplicates

jstrachan:
We merge presubmits or postsubmits by name when using shared configurations and in-repo configurations; or when using in-repo configurations on a particular branch versus master.

When merging triggers we need to make sure we combine them together if they have the same name; otherwise we end up with a configuration validation error with duplicate presubmits/postsubmits by name or for the same branch on a post submit

Without this fix, its quite easy to get lighthouse to fail to process a webhook if it has in-repo configurations as it ends up creating a duplicate Presubmit/Postsubmit and then refuses to process the webhook

### DIFF
--- a/pkg/triggerconfig/inrepo/calculate_test.go
+++ b/pkg/triggerconfig/inrepo/calculate_test.go
@@ -110,6 +110,6 @@ func TestTriggersInBranchMergeToMaster(t *testing.T) {
 	assert.Contains(t, presubmitNames, "lint", "presubmits for repo %s", fullName)
 	assert.Contains(t, presubmitNames, "newthingy", "presubmits for repo %s", fullName)
 
-	assert.Len(t, cfg.Postsubmits[fullName], 0, "postsubmits for repo %s", fullName)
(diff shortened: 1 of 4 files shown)


abayer:
I'm not entirely sure what this is doing - we can have multiple postsubmits as long as they have different names.

jstrachan:
@abayer sorry for the cryptic comments/PR. 

I've tried updating the above comment to be a bit more clear. Basically the logic when merging either shared + in repo pre/post submits; or master + custom branch in repo pre/post submits we were not merging by name but just appending everything together into a list; so it was very easy to create validation errors if the pre/post submits for a repository had duplicates for the same name/branch.

This change should not have any effect on folks not using in-repo configuration at all; but should help avoid LH failing unnecessarily with "duplicate pre/post submit errors"

jstrachan:
incidentally I've squashed the commits + also modified the postsubmit logic to be polymorphic with presubmits; the merge is always based on names. So shared + in repo (or master + custom branch in repo) merging always combines pre/post submits together if they are the same name; otherwise they are added to the list

PR #1078, opened 2020-09-24
test: Add BDD tests for Lighthouse + Jenkins

### LINKED ISSUE #1048: Create end to end test for Jenkins integration
The gist of the work needed is:

- Create a new Jenkins X context, aka `jenkins-x-jenkins.yml`
    - difference to the others is that in the end it calls something like `bdd/jenkins/ci.sh`
- Create `bdd/jenkins/ci.sh`
    - create the test cluster
    - install Jenkins 
    - install Lighthouse
- Create jenkins_test.go
    - create a test repo
    - create the Jenkins  test pipeline 
    - configure Lighhouse for use of Jenkins and the test repo
    - make a change to the git repo
    - verify Jenkins jobs runs

See also https://github.com/jenkins-x/lighthouse/pull/1033#issuecomment-691145435

hferentschik:
fixes #1048


hferentschik:
@abayer, I think it's looking good now. The tests creates the Lighthouse + Jenkins configuration, creates a successful and an failing PR and in each case verifies against the Jenkins build (via build log). 

Before merging though, PR #1088 needs to merge. Also what was the dance again regarding enabling the new context. First I change this PR to not use the gitlab context and then we merge and afterwards we add the new context to the prod config so that the new context will also build moving forward, right?

Any other thoughts on this?

PR #1080, opened 2020-09-28
fix: run as an unprivileged user

mateuszpodraza:
This PR changes Dockerfiles to that all the different bits of lighthouse can run as a non-root user.


abayer:
/lgtm

Thanks, looks good!

PR #1081, opened 2020-09-28
fix: use numeric user for the image

mateuszpodraza:
While #1080 did what it was supposed to do (run as a non-root user), it didn't solve the problem I originally wanted to solve - have lighthouse run under a restrictive PodSecurityPolicy.

It currently fails with `container has runAsNonRoot and image has non-numeric user`.
From the official docs for MustRunAsNonRoot:
> Requires that the pod be submitted with a non-zero runAsUser or have the USER directive defined (**using a numeric UID**) in the image


abayer:
/retest

Random test failure - rerunning.

PR #1092, opened 2020-09-30
chore: Set the default replica count for webhooks to 1

### LINKED ISSUE #1091: Per default have only one replica of the Lighthouse webhook handler
Currently the default for the webhook deployment is 2:

```
webhooks:
  # webhooks.replicaCount -- Number of replicas
  replicaCount: 2
```

I think it should be ok to have only a single replica as for the other Lighthouse deployments. I don't see a driving reason for this replica count.

hferentschik:
fixes #1091


eddycharly:
I hope it's ok to change the default replicas count from the jenkins-x perspective.

/lgtm
/approve

PR #1100, opened 2020-10-03
docs: explain how to swap deployments using telepresence

yelhouti:


yelhouti:
Hi @abayer :).
I only changed a README :p could you send the logs here so I can see why it's complaining? thanks

PR #1102, opened 2020-10-05
chore: K8s upgrade 0.19.2

### LINKED ISSUE #1101: Upgrade go module dependencies for Kubernetes (v0.19.2) to align with all Jenkins X repos
We are aligning Kubernetes dependencies across all Jenkins X libraries and plugins to 0.19.2.  This also includes an upgrade of client-go which brings in some breaking API changes.

rawlingsj:
Included in this change set is the use of a Tekton fork while we wait for their dependency upgrade.  I don't think there's any risk that they don't upgrade, it's just a matter of when, which we already have a fork ready to PR.  Details of that can be tracked here https://github.com/tektoncd/pipeline/issues/3324.  As the Tekton fork doesn't include any functionality changes and is purely a dependency update hopefully this PR will be fine.

Fixes #1101


hferentschik:
Personally, I'd like to see Tekton update first. If you switch to a fork it will be harder to upgrade to interim releases of Tekton until 0.19 is reached.

rawlingsj:
It's pretty easy to apply the 0.19.2 updates periodically to Tekton using the steps here https://github.com/tektoncd/pipeline/issues/3324 , I'm happy to do that once a week or however regularly you think we need to update to the latest Tekton APIs.  But putting LH on the latest would be great as it aligns all Jenkins X repos plus it means LH is using the newer Kubernetes APIs sooner.

abayer:
Can this wait a month until Tekton 0.18 releases? That will be switching to k8s 0.18 with the new function signatures.

rawlingsj:
We've been aligning with 0.19.2 rather than 0.18 as a few other projects including Kubernetes ones seem to be on that version.

rawlingsj:
We've just finishing moving all Jenkins X 3.x go based repos over and it's a nice clean go mod file now too so easier to maintain and track upstream.

rawlingsj:
It's totally up to you, the hard work has been done in this PR and BDD is passing which means LH can get on the latest.  The Jenkins X 3.x plugins can use a branch of LH so they're not blocked but it does mean you'd need to redo this PR again in the future.  Up to you, if you want to close it that's fine.

hferentschik:
>  The Jenkins X 3.x plugins can use a branch of LH so they're not blocked but it does mean you'd need to redo this PR again in the future. Up to you, if you want to close it that's fine.

OOI, are you using Lighthouse as a compile time dependency? It's not just a component you deploy?

rawlingsj:
> OOI, are you using Lighthouse as a compile time dependency? 

Yes correct, one or two jx plugins, for example https://github.com/jenkins-x/jx-gitops/blob/2d571b66129387830e3f91c9561ed0727068e7ce/go.mod#L16

PR #1117, opened 2020-10-16
fix: lets remove the hard coding of master branch

### LINKED ISSUE #1110: Initial .lighthouse config check need to be done on base branch on which PR/release is made not master branch all times.
With new tekton catalog pipeline jx3 version build is not triggering due to following reason on lighthouse/webhook controller

`{"Branch":"qa","Clone":"https://github.com/org/node-service.git","ID":"143455","Link":"https://github.com/org/node-service","Name":"node-service","Namespace":"org","Webhook":"pull_request","author":"author","error":"failed to calculate in repo config: failed to create trigger config from local source for repo org/node-service ref master: failed to find any lighthouse configuration files in repo org/node-service at sha master: Not Found","level":"error","msg":"Error creating agent for PullRequestEvent.","org":"org","plugin":"size","pr":599,"repo":"node-service","time":"2020-10-12T05:04:34Z","url":"https://github.com/org/node-service/pull/99"}`

Seems jx3 lighthouse trigger always initially checks **master** branch for presence of .lighthouse config which shouldn't be done as per me, instead it should validate the presence of the same config only on the branch in which event is made(base branch of pull and release branch of release).

jstrachan:
so that the in-repo logic works with whatever the main branch is

fixes #1110

### DIFF
--- a/pkg/triggerconfig/inrepo/calculate.go
+++ b/pkg/triggerconfig/inrepo/calculate.go
@@ -47,7 +47,7 @@ func Generate(scmClient scmProviderClient, sharedConfig *config.Config, sharedPl
 
 	eventRef = strings.TrimPrefix(eventRef, "refs/heads/")
 	eventRef = strings.TrimPrefix(eventRef, "refs/tags/")
-	if eventRef != mainBranch {
+	if eventRef != mainBranch && eventRef != "" {
 		refs = append(refs, eventRef)
 	}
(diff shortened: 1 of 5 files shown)


jstrachan:
I've tested this fix in the jx3 BDD tests and we've been using it in production for a few days and it all seems good for handing both `main` and `master` based repos now

PR #1121, opened 2020-10-21
fix: bug in merge logic

jstrachan:
it looks like we were incorrectly updating the existing presubmit/postsubmits when merging - we were using the index variable from the wrong for loop ;)


abayer:
/retest

@hferentschik - we should eventually figure out why the jenkins context is flaky and fix it. =)

PR #1131, opened 2020-10-27
fix: log failure raison since not possible to sent to gitlab

yelhouti:
Signed-off-by: Youssef El Houti <youssef.elhouti@gmail.com>


jstrachan:
I'm cool with this workaround for now (to avoid barfing with keeper state changes on gitlab).

You mentioned `GitLab doesn't like updating status description without a state change` - I just wondered if we could make a state change for gitlab so that we can update the status. I guess that would mean some kind of commit or is there another way of making a state change on gitlab so that we can update the status? (e.g. a comment/label or anything else?)

yelhouti:
I guess we can switch tu running then back to pending, I will try that later

PR #1132, opened 2020-10-27
fix: use assign PR in plugins when working with PR instead of assign issue

yelhouti:


jstrachan:
@yelhouti looks like `make vet` fails...
```
vet: pkg/plugins/assign/assign_test.go:428:75: cannot use fc (variable of type *fakeClient) as scmProviderClient value in argument to newAssignHandler: missing method AssignPR
```

yelhouti:
last test failing is because of:
https://github.com/jenkins-x/go-scm/pull/203
please merge this and rebase my PR for this test to pass

PR #1146, opened 2020-11-03
feat: handle custom trigger commands for a deployment

### LINKED ISSUE #1145: allow webhook to handle a custom trigger command so that a specific deployment can support custom chat prefixes
For a lighthouse webhook deployment we have a usecase where we want to accept specific customised trigger command. 

i.e. with the default test command (as in `/test foo`) which is here https://github.com/jenkins-x/lighthouse/blob/a740deb88753c58780ab966fb82db7252646663e/pkg/plugins/trigger/trigger.go#L56-L65 along with the supporting regex we'd like to open it up to support extending that to include a custom command so folks can configure triggers such as
```
/foo mycoolthing
```
or a more specific case
```
/jx lint
```
The top level command seems needed after debugging it, this also means there is less risk of clashing with other configured plugins.

rawlingsj:
deployments can configure the ability to allow custom lighthouse triggers using their own unique chat prefix,
for example extending the default `/test` trigger prefix let them specify the helm chart value

```
webhooks.customDeploymentTriggerCommand=foo
````

which means they can also use their own custom trigger /foo mycoolthing

fixes #1145


rawlingsj:
Is the jenkins context expected to pass normally?

jstrachan:
@rawlingsj its a bit flaky

PR #1164, opened 2020-11-18
fix: reduce the number of github API calls

jstrachan:
lets setup the agent when using in-repo calls upfront rather than repeating for each plugin


rawlingsj:
we've been running this on Jenkins X own infra for a couple of weeks and not seen one api limit hit since so can verify this is a big improvement.

PR #1166, opened 2020-11-26
fix(chart): foghorn is missing the HMAC token

vbehar:
which results in activity records events sent to external plugin not being signed with the right token


vbehar:
/assign @jstrachan 
thanks!

PR #1179, opened 2020-12-20
fix: build pull request broken when run_if_changed is set

wingyplus:
The problem found when user specified `run_if_changed` in triggers.yaml. It can make lighthouse crash with nil pointer dereference panic because `run_if_changed` required to compile regular expression before use it. The workaround for this problem fixed in https://github.com/jenkins-x/lighthouse/commit/ff0c2cb55ba0b184a4059fe5f0877a2a33f2f3de. After dig down into the problem, I found that configuration doesn't initialize before validate (ref this function call https://github.com/wingyplus/lighthouse/blob/25e654ab141d8280b97c4aa9c7e9279a0d35fe12/pkg/triggerconfig/merge/merge.go#L87). So try fix it by add `cfg.Init` before `cfg.Validate` calls. 

After adding Init function calls. I found that tests was fail due to `rerun_command` doesn't be the subset of `trigger` and field name is incorrect, such as `rerunCommand` (should be `rerun_command`) or `alwaysRun`( should be `always_run`). This changes correct the field name and add `retest` rerun_command to be the subset of `trigger` to make tests pass.

### DIFF
--- a/pkg/triggerconfig/inrepo/test_data/myorg/branchtest/refs/master/.lighthouse/triggers/triggers.yaml
+++ b/pkg/triggerconfig/inrepo/test_data/myorg/branchtest/refs/master/.lighthouse/triggers/triggers.yaml
@@ -4,10 +4,10 @@ spec:
   presubmits:
   - name: lint
     context: "lint"
-    alwaysRun: true
+    always_run: true
     optional: false
-    trigger: "/lint"
-    rerunCommand: "/relint"
+    trigger: "(?:/lint|/relint)"
+    rerun_command: "/relint"
     agent: tekton-pipeline
   postsubmits:
   - name: release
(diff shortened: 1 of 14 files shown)


wingyplus:
@abayer I cannot access to the  details. I'm not sure why jenkins pipeline was failed. :(

abayer:
/retest

@wingyplus Sadly, the jenkins context is flaky - I'll just keep rerunning it.

wingyplus:
> /retest
> 
> @wingyplus Sadly, the jenkins context is flaky - I'll just keep rerunning it.

T T

abayer:
/retest

I just fixed the problem with the other failures - rerunning.

wingyplus:
@abayer Thanks for accepted patch. I've a question, what's override jenkins means?

hervelemeur:
@wingyplus this chatops command is used to override the "jenkins" pipeline result, allowing to merge the PR even if it failed.

Issue #1191, opened 2021-01-01
webhook pod panic


cameronbraid:
```
lighthouse-webhooks-8576c4bc44-xzrst lighthouse-webhooks 2021/01/01 13:37:47 http: panic serving 10.244.0.6:35566: runtime error: invalid memory address or nil pointer dereference
lighthouse-webhooks-8576c4bc44-xzrst lighthouse-webhooks goroutine 77 [running]:
lighthouse-webhooks-8576c4bc44-xzrst lighthouse-webhooks net/http.(*conn).serve.func1(0xc000332460)
lighthouse-webhooks-8576c4bc44-xzrst lighthouse-webhooks        /usr/local/go/src/net/http/server.go:1767 +0x139
lighthouse-webhooks-8576c4bc44-xzrst lighthouse-webhooks panic(0x1743920, 0x2879060)
lighthouse-webhooks-8576c4bc44-xzrst lighthouse-webhooks        /usr/local/go/src/runtime/panic.go:679 +0x1b2
lighthouse-webhooks-8576c4bc44-xzrst lighthouse-webhooks github.com/jenkins-x/lighthouse/pkg/filebrowser.(*gitFileBrowser).withRepoClient(0xc00064cde0, 0xc0004f74b8, 0x6, 0xc0004f7640, 0x10, 0x1965225, 0x0, 0xc0004c3f70, 0x203000, 0x203000)
lighthouse-webhooks-8576c4bc44-xzrst lighthouse-webhooks        /workspace/source/pkg/filebrowser/git_file_browser.go:98 +0x378
lighthouse-webhooks-8576c4bc44-xzrst lighthouse-webhooks github.com/jenkins-x/lighthouse/pkg/filebrowser.(*gitFileBrowser).ListFiles(0xc00064cde0, 0xc0004f74b8, 0x6, 0xc0004f7640, 0x10, 0x196e105, 0xb, 0x1965225, 0x0, 0x0, ...)
lighthouse-webhooks-8576c4bc44-xzrst lighthouse-webhooks        /workspace/source/pkg/filebrowser/git_file_browser.go:46 +0xfb
[...]

PR #1192, opened 2021-01-01
fix: return the err to prevent a panic below when client.repoClient is nil

cameronbraid:
fixes https://github.com/jenkins-x/lighthouse/issues/1191


jstrachan:
/override jenkins

the test is flaky

PR #1194, opened 2021-01-04
fix: convert old incorrect configuration

jstrachan:
lots of in-repo triggers have incorrectly used `/test` and `/retest` and `/lint` and `/relint` for `trigger` and `rerun_command` due to the bug where we didn't used to properly validate JobConfig for in repo.

to make it easy to transition to the latest lighthouse lets add a little migrator function to switch those bad configurations to their correct values so that lighthouse still works if its upgraded for existing triggers in Jenkins X


rawlingsj:
great - this avoids breaking existing users inrepo config when upgrading to latest lighthouse
/lgtm

PR #1196, opened 2021-01-04
feat: allow user to specify pod annotations for gc job

patrickleet:
Signed-off-by: Patrick Lee Scott <pat@patscott.io>


patrickleet:
My use case for this is disabling istio injection when it's enabled on the namespace

PR #1216, opened 2021-01-20
fix: keeper status case insensitive labels

patrickleet:
Signed-off-by: Patrick Lee Scott <pat@patscott.io>


patrickleet:
/cc @abayer 

Thanks for merging the last two - still didn't work though so had to dig deeper.

Added a new test case that failed before the change as well.

abayer:
/ok-to-test
/lgtm

d'oh, I shoulda thought of that too! =)

PR #1229, opened 2021-02-02
chore: tone down logging to debug level

jstrachan:
to reduce log noise


jstrachan:
/override ghe as our GHE server is currently down

jstrachan:
/override ghe 
as our GHE server is currently down

PR #1233, opened 2021-02-08
fix: prometheus support for keeper

vbehar:
- ensure prometheus metrics are exposed if required
- fix chart to allow users to add the prometheus scrape annotations on the pods


jstrachan:
/override ghe 

our GHE server is down

vbehar:
thanks, I wanted to ask you what was the issue ;-)

Issue #1234, opened 2021-02-10
bug in the uses resolver when overriding an env of a 'shared' PipelineRun


cameronbraid:
I have a **release** pipeline that is ending up with env vars as defined in the **pr** pipeline. 

 Both pr and release pipelines use a step like the following

**PR does this :**
```
        - name: nitro-webpack-publish
          image: uses:.lighthouse/jenkins-x/nitro-build.yaml
          env: 
            - name: REMOTE_PATH
              value: /public/nitro-pr/
```
**Release does this :**
```
        - name: nitro-webpack-publish
          image: uses:.lighthouse/jenkins-x/nitro-build.yaml
```

But the release pipeline ends up with `REMOTE_PATH=/public/nitro-pr/`  where as the step definition  in `.lighthouse/jenkins-x/nitro-build.yaml` is
```
        - name: nitro-webpack-publish
          image: rclone/rclone
          envFrom:
            - secretRef:
                name: minio-auth
          env:
            - name: LOCAL_PATH
              value: ./webpack/
            - name: S3_ENDPOINT
              value: http://minio.minio.svc.cluster.local:9000
            - name: REMOTE_PATH
              value: /drivenow-shared-files/webdata/nitro/
          script: |
            #!/usr/bin/env sh
            source /workspace/source/.jx/variables.sh
            set -x -e
            rclone \
              --s3-access-key-id="${AWS_ACCESS_KEY}" \
              --s3-secret-access-key="${AWS_SECRET_KEY}" \
              --s3-endpoint="${S3_ENDPOINT}" \
              --no-traverse \
              --progress \
              copy \
              "${LOCAL_PATH}" \
              ":s3:${REMOTE_PATH}"
```

I suspect all that is required is the uses resolver to clone the imported struct that is resolved from .lighthouse/jenkins-x/nitro-build.yaml

cameronbraid:
PR to come

PR #1235, opened 2021-02-10
fix: in uses resolver, clone pipelinerun when returning from the cache to avoid changes persistng between evaluations

cameronbraid:
The test runner was changed to evaluate multiple source/expected runs

e.g. 
source.yaml compared to expected.yaml
source2.yaml compared to expected2.yaml

Fix was a `pr.DeepClone()`

Fixes https://github.com/jenkins-x/lighthouse/issues/1234


jstrachan:
/override Pull Request Build / MarkdownLinkCheck (pull_request)

PR #1238, opened 2021-02-10
chore: lets allow the hmac token to be disabled

jstrachan:
for bitbucketcloud support


jstrachan:
/override jenkins
its a very flaky test

PR #1252, opened 2021-03-04
fix: enable LocalFileResolve in UsesResolver

wingyplus:


tomhobson (changes requested):
Hey!

Can you give some information about why this is required? Or link it to an issue if it's a resolution for a problem?

wingyplus:
Unfortunately, I dont remember why should be enable since this patch opened 1 year ago. 😢 

Feel free to close it if it’s not make sense.

tomhobson:
Closing as I'm not sure on the purpose of this. 

@wingyplus thanks for the contribution though!

PR #1291, opened 2021-04-22
docs: Docs update

mariadb-JeffBachtel:
- Documents that `webhooks.ingress.hosts` is required when ingress is enabled (otherwise chart fails to deploy)
- Updates helm-docs to 1.5.0
- Uses new, shorter helm-docs syntax in `values.yaml`


LinuxSuRen (approved):
LGTM

msvticket:
No response from the developer

Issue #1306, opened 2021-05-12
Cache issue with uses resolver when a PR branch and Master branch both define a shared task.yaml which is used via uses: syntax


cameronbraid:
The PR to follow will include an example of when this occurs, and will save me from outlining it here

cameronbraid:
Fixed in

PR #1307, opened 2021-05-12
fix: fix uses resolver using incorrect version of uses:task.yaml

cameronbraid:
Test case in TestUsesInPRBranch, test data in pkg/triggerconfig/inrepo/test_data/myorg/uses-cache-issue

The fix was to include the SHA in the uses resolver cache key so that a different entry exists for master and the pr branch for a given path.


cameronbraid:
Fixes https://github.com/jenkins-x/lighthouse/issues/1306

PR #1316, opened 2021-05-28
chore: move from google container registry to github packages

rawlingsj:


LinuxSuRen:
Currently, there's no connection between the image and GitHub repo. Please see the following screenshot:

image

So, I'm wondering if it's possible to add a label to the images. For example: `org.opencontainers.image.source = "https://github.com/jenkins-x/lighthouse"`

Issue #10, opened 2019-07-05
Load the plugin configuration from the Schedule CRD


jstrachan:
we need to load the `Schedule` configuration for the current SourceRepository when a webhook comes in to know what plugins/configuration to use. Right now in the PoC we're using an environment variable ;)

jstrachan:
we are currently reusing the Prow config maps; so this one can be left on the back log for a while I guess; its not urgent

jstrachan:
not totally sure this is required - we load the `config / plugins ConfigMaps` which are generated by `jx boott` for both prow + lighthouse

PR #1388, opened 2021-11-21
fix: GIT_USER

matiasgonzalocalvo:
Error from server (BadRequest): error when creating "config-root/namespaces/jx/lighthouse/lighthouse-foghorn-deploy.yaml": Deployment in version "v1" cannot be handled as a Deployment: v1.Deployment.Spec: v1.DeploymentSpec.Template: v1.PodTemplateSpec.Spec: v1.PodSpec.Containers: []v1.Container: v1.Container.Env: []v1.EnvVar: v1.EnvVar.Value: ReadString: expects " or n, but found 3, error found in #10 byte of ...|,"value":36821702},{|..., bigger context ...|"https://github.com"},{"name":"GIT_USER","value":36821702},{"name":"GIT_TOKEN","valueFrom":{"secretK|...

### DIFF
--- a/charts/lighthouse/templates/foghorn-deployment.yaml
+++ b/charts/lighthouse/templates/foghorn-deployment.yaml
@@ -36,7 +36,7 @@ spec:
             value: "/secrets/githubapp/tokens"
 {{- else }}
           - name: "GIT_USER"
-            value: {{ .Values.user }}
+            value: "{{ .Values.user }}"
           - name: "GIT_TOKEN"
             valueFrom:
               secretKeyRef:


charts/lighthouse/templates/foghorn-deployment.yaml
```diff
@@ -36,7 +36,7 @@ spec:
             value: "/secrets/githubapp/tokens"
 {{- else }}
           - name: "GIT_USER"
-            value: {{ .Values.user }}
+            value: "{{ .Values.user }}"
```
ankitm123, line 39: Are u sure this is the real issue?

tomhobson:
Closing as it's an inactive PR, feel free to reopen if required!

PR #1392, opened 2021-12-01
feat: show error message when pipeline syntax is wrong

ankitm123:


nicolaerosia:
This would be helpful!

PR #1395, opened 2021-12-09
feat: Use OWNERS_ALIASES files from other repos

### LINKED ISSUE #1382: Feature request: Share aliases between repositories
As a fall back for aliases missing from the OWNER_ALIASES file I propose support for reading OWNER_ALIASES files from other repositories. 

This is practical when teams have responsibility for several repositories.

msvticket:
For organizations with many repositories it would make sense to refer to other repos to find OWNERS_ALIASES files in.

This PR implements the added field foreignAliases.

So an example OWNERS_ALIASES could look like
```
aliases:
  best-approvers:
  - you
  - me
foreignAliases:
- name: a-repo
- org: another-org
  name: the-repo
- org: yaorg
  name: yarepo
  ref: special-branch
```

org defaults to current organization
ref default to HEAD

If an alias is present in several sources the members are merged.

Fixes #1382

### DIFF
--- a/pkg/repoowners/repoowners.go
+++ b/pkg/repoowners/repoowners.go
@@ -77,6 +77,7 @@ type FullConfig struct {
 type scmProviderClient interface {
 	ListCollaborators(org, repo string) ([]scm.User, error)
 	GetRef(org, repo, ref string) (string, error)
+	GetFile(org, repo, filepath, ref string) ([]byte, error)
 }
 
 type cacheEntry struct {
@@ -132,6 +133,19 @@ func NewClient(
 
 // RepoAliases defines groups of people to be used in OWNERS files
 type RepoAliases map[string]sets.String
+
+// ForeignAlias contains a
+type ForeignAlias struct {
+	Org  string `json:"org"`
+	Repo string `json:"name"`
+	Ref  string `json:"ref,omitempty"`
+}
+
+// OwnerAliases contains the parsed OWNER_ALIASES config.
+type OwnerAliases struct {
+	Aliases        map[string][]string `json:"aliases,omitempty"`
+	ForeignAliases []ForeignAlias      `json:"foreignAliases,omitempty"`
+}
 
 // RepoOwner is an interface to work with repoowners
 type RepoOwner interface {
@@ -192,7 +206,7 @@ func (c *Client) LoadRepoAliases(org, repo, base string) (RepoAliases, error) {
 			return nil, err
 		}
 
-		entry.aliases = loadAliasesFrom(gitRepo.Dir, log)
+		entry.aliases = c.loadAliasesFrom(gitRepo.Dir, log, org)
 		entry.sha = sha
 		c.cache[fullName] = entry
 	}
@@ -228,7 +242,7 @@ func (c *Client) LoadRepoOwners(org, repo, base string) (RepoOwner, error) {
 
 		if entry.aliases == nil || entry.sha != sha {
 			// aliases must be loaded
-			entry.aliases = loadAliasesFrom(gitRepo.Dir, log)
+			entry.aliases = c.loadAliasesFrom(gitRepo.Dir, log, org)
 		}
 
 		excludeConfig := c.config.OwnersDirExcludes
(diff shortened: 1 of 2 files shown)


pkg/repoowners/repoowners.go
```diff
@@ -77,6 +77,7 @@ type FullConfig struct {
 type scmProviderClient interface {
 	ListCollaborators(org, repo string) ([]scm.User, error)
 	GetRef(org, repo, ref string) (string, error)
+	GetFile(string, string, string, string) ([]byte, error)
```
ankitm123, line 80: Add a more descriptive name to the args instead of (string, string, string, string), just like ListCollaborators and GetRef ...

pkg/repoowners/repoowners.go
```diff
@@ -304,22 +318,49 @@ func loadAliasesFrom(baseDir string, log *logrus.Entry) RepoAliases {
...
+		log.WithError(err).Errorf("Failed to unmarshal aliases from %q.", path)
 	}
 	log.Infof("Loaded %d aliases from %q.", len(result), path)
 	return result
 }
 
+func (c *Client) parseAliases(result RepoAliases, b []byte, log *logrus.Entry, org string) (err error) {
+	config := &OwnerAliases{}
+	if err = yaml.Unmarshal(b, config); err != nil {
+		return
```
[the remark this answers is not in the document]
msvticket: OK

pkg/repoowners/repoowners.go
msvticket: OK. I just copied from pkg/scmprovider/client.go:31 ...

pkg/repoowners/repoowners.go
```diff
@@ -304,22 +318,50 @@ func loadAliasesFrom(baseDir string, log *logrus.Entry) RepoAliases {
...
+	if config.ForeignAliases != nil {
+		for _, r := range config.ForeignAliases {
+			if r.Ref == "" {
+				r.Ref = "HEAD"
+			}
+			if r.Org == "" {
+				r.Org = org
+			}
+			b, err = c.spc.GetFile(r.Org, r.Repo, aliasesFileName, r.Ref)
+			if err == nil {
```
ankitm123, line 345: Return error using err != nil? This error is not being handled, same with `c.parseAliases`, we should return those errors before we proceed in the code imo ...

pkg/repoowners/repoowners.go
```diff
@@ -133,6 +134,19 @@ func NewClient(
 // RepoAliases defines groups of people to be used in OWNERS files
 type RepoAliases map[string]sets.String
 
+// ForeignAlias contains a
+type ForeignAlias struct {
+	Org  string `json:"org"`
+	Repo string `json:"name"`
+	Ref  string `json:"ref,omitempty"`
```
ankitm123, line 141: Should this also be an option? I thought we always want to take the value from master/main branch, right? Curious ...

ankitm123 (commented):
Over all looks good, just a few minor comments, thanks for adding a test case 💯

pkg/repoowners/repoowners.go
msvticket: Well, I need to provide a reference to GetFile anyway so I thought I could as well make it configurable.

pkg/repoowners/repoowners.go
msvticket: The error is returned in the end of the function. But we don't want to abort the processing because of an error.
The only drawback with the current code is that if there are several errors only the last will be reported. But I think that is OK.

ankitm123: can we try to aggregate the errors then? we dont want to lose the errors, else the end user will have a hard time knowing the exact issue.

pkg/repoowners/repoowners.go
ankitm123: So trying to understand, if I open a PR in the repo and edit the alias, and it is still not merged into main/master, is it possible for me to have approving rights?

msvticket: No

pkg/repoowners/repoowners.go
msvticket: We could. But since this can only happen in the case several repos are specified and there are failures in accessing several of them it is really an edge case that I don't know if it's really worth adding complexity to solve.

ankitm123: I also wonder if it will get flagged by errcheck when linting is enabled on lighthouse in the next few weeks 🤔 
@rawlingsj wdyt?

ankitm123: I think it's fine for now, I will see what happens when we add linting to lighthouse.

ankitm123 (approved):

pkg/repoowners/repoowners.go
ankitm123: I do see that we handle the error from find at other places, approved, but putting a hold for @rawlingsj's review.

rawlingsj: yeah that all sounds good to me

PR #1399, opened 2021-12-17
feat: Support poller release and pull request periods

dippynark:
This PR allows the period between release and pull request polls to be specified separately.

### DIFF
--- a/cmd/poller/main.go
+++ b/cmd/poller/main.go
@@ -47,6 +47,7 @@ type options struct {
 	disablePollRelease     bool
 	disablePollPullRequest bool
 	pollPeriod             time.Duration
+	pollReleasePeriod      time.Duration
 }
 
 func (o *options) Validate() error {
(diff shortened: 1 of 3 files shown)


cmd/poller/main.go
```diff
@@ -94,6 +95,19 @@ func gatherOptions(fs *flag.FlagSet, args ...string) options {
 	return o
 }
 
+func parseEnvPollPeriod(env string, defaultPollPeriod time.Duration) time.Duration {
+	text := os.Getenv(env)
+	if text != "" {
+		d, err := time.ParseDuration(text)
+		if err != nil {
+			logrus.WithError(err).Warn(fmt.Sprintf("invalid %s value", env))
```
eddycharly, line 103: Can we add a more detailed message ?
Actual value and expected format would be great.

eddycharly: Maybe using something like WithField ?

dippynark: Added some improvements, hopefully that covers it

eddycharly: Looks good, thanks !

PR #1401, opened 2021-12-17
feat: Add ingress class name and upgrade to stable ingress

slimm609:
Add support for ingress class name and stable ingress if available

Signed-off-by: Brian Davis <dbrian@vmware.com>

### DIFF
--- a/charts/lighthouse/templates/webhooks-ingress.yaml
+++ b/charts/lighthouse/templates/webhooks-ingress.yaml
@@ -1,10 +1,7 @@
-{{- $apiAll := .Capabilities.APIVersions | toStrings -}}
-{{- $ingStableVersion := 19 | int -}}
-
 {{- if .Values.webhooks.ingress.enabled }}
-{{- if ge ($.Capabilities.KubeVersion.Minor | int) $ingStableVersion }}
+{{- if .Capabilities.APIVersions.Has "networking.k8s.io/v1" }}
 apiVersion: networking.k8s.io/v1
-{{- else if has "networking.k8s.io/v1beta1" $apiAll }}
+{{- else if .Capabilities.APIVersions.Has "networking.k8s.io/v1beta1" }}
 apiVersion: networking.k8s.io/v1beta1
 {{- else }}
 apiVersion: extensions/v1beta1
@@ -26,7 +23,7 @@ spec:
     http:
       paths:
         - backend:
-        {{- if ge ($.Capabilities.KubeVersion.Minor | int) $ingStableVersion }}
+        {{- if $.Capabilities.APIVersions.Has "networking.k8s.io/v1" }}
             service:
               name: {{ default (include "webhooks.name" $) $.Values.webhooks.serviceName }}
               port:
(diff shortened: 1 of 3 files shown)


slimm609:
Updated to add README line

charts/lighthouse/templates/webhooks-ingress.yaml
```diff
@@ -1,5 +1,14 @@
+{{- $apiAll := .Capabilities.APIVersions | toStrings -}}
```
eddycharly, line 1: Can’t we use .Capabilities.APIVersions.Has here ?

charts/lighthouse/templates/webhooks-ingress.yaml
```diff
@@ -8,13 +17,25 @@ metadata:
...
 spec:
+  {{- if .Values.webhooks.ingress.ingressClassName }}
+  ingressClassName: {{ .Values.webhooks.ingress.ingressClassName }}
+  {{- end }}
   rules:
   {{- range .Values.webhooks.ingress.hosts }}
   - host: {{ . | quote }}
     http:
       paths:
         - backend:
```
eddycharly, line 28: Can we keep the logic at a single place ?

slimm609: IngressClassName was added in 1.18 and was not part of the stable vs beta so it doesn't pin to the apiVersion directly.

charts/lighthouse/templates/webhooks-ingress.yaml
slimm609: testing

charts/lighthouse/templates/webhooks-ingress.yaml
eddycharly: Maybe just add `{{- if semverCompare ">=1.18-0" .Capabilities.KubeVersion.Version }}` ?

charts/lighthouse/templates/webhooks-ingress.yaml
slimm609: It works here but it doesn't below inside the backend.  Trying a few things to see if I can get it to accept it

```bash
<.Capabilities.APIVersions.Has>: can't evaluate field Capabilities in type interface {}
```

eddycharly: Yeah, you need to prefix with `$` because you're in a range.

charts/lighthouse/templates/webhooks-ingress.yaml
slimm609: You would still need to set a value though, otherwise it would be blank

eddycharly: Just set `ingressClassName: null` in `values.yaml`

charts/lighthouse/templates/webhooks-ingress.yaml
slimm609: Updated

charts/lighthouse/templates/webhooks-ingress.yaml
slimm609: updated as well

PR #1403, opened 2021-12-17
fix: do not include tekton roles when disabled

slimm609:
- do not include tekton roles when tekton engine is disabled

Signed-off-by: Brian Davis <dbrian@vmware.com>


eddycharly:
Seems reasonable to me.
/ok-to-test

PR #1411, opened 2022-01-07
feat: allow users in owners file outside of org to trigger pipelines

### LINKED ISSUE #1313: Allow Pull Requests to be automatically tested when the author is in the OWNERS file
Currently, PRs are only automatically tested when an author is in the organization or if it's a Jenkins-X bot. 

My use case would be to allow GitHub's depandabot or other bots PR's to be automatically tested, provided they are added to the OWNERS file, as commenting `/ok-to-test` for each of its PR's is proving to be a bit of a hassle.

ankitm123:

fixes #1313

### DIFF
--- a/pkg/plugins/trigger/pull-request_test.go
+++ b/pkg/plugins/trigger/pull-request_test.go
@@ -25,9 +25,71 @@ import (
 	"github.com/jenkins-x/lighthouse/pkg/labels"
 	"github.com/jenkins-x/lighthouse/pkg/launcher/fake"
 	"github.com/jenkins-x/lighthouse/pkg/plugins"
+	"github.com/jenkins-x/lighthouse/pkg/repoowners"
 	fake2 "github.com/jenkins-x/lighthouse/pkg/scmprovider/fake"
 	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/util/sets"
 )
+
+type fakeOwnersClient struct{}
+
+func (foc fakeOwnersClient) LoadRepoOwners(org, repo, base string) (repoowners.RepoOwner, error) {
+	return fakeRepoOwners{
+		fakeRepo: fakeRepo{
+			approvers: map[string]sets.String{
+				"":        sets.NewString("jx-bot"),
+				"CHANGED": sets.NewString("jx-bot"),
+			},
+			leafApprovers:  map[string]sets.String{},
+			approverOwners: map[string]string{},
+		},
+	}, nil
+}
+
+type fakeRepo struct {
+	approvers, leafApprovers map[string]sets.String
+	approverOwners           map[string]string
+}
+
+func (fr fakeRepo) Approvers(path string) sets.String {
+	return fr.approvers[path]
+}
+
+func (fr fakeRepo) LeafApprovers(path string) sets.String {
+	return fr.leafApprovers[path]
+}
+
+func (fr fakeRepo) FindApproverOwnersForFile(path string) string {
+	return fr.approverOwners[path]
+}
+
+func (fr fakeRepo) IsNoParentOwners(path string) bool {
+	return false
+}
+
+type fakeRepoOwners struct {
+	fakeRepo
+}
+
+func (fro fakeRepoOwners) FindLabelsForFile(path string) sets.String {
+	return sets.NewString()
+}
+
+func (fro fakeRepoOwners) FindReviewersOwnersForFile(path string) string {
+	return ""
+}
+
+func (fro fakeRepoOwners) LeafReviewers(path string) sets.String {
+	return sets.NewString()
+}
+
+func (fro fakeRepoOwners) Reviewers(path string) sets.String {
+	return sets.NewString()
+}
+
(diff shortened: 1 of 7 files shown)


pkg/plugins/trigger/pull-request_test.go
```diff
@@ -25,17 +25,78 @@ import (
 	"github.com/jenkins-x/lighthouse/pkg/labels"
 	"github.com/jenkins-x/lighthouse/pkg/launcher/fake"
 	"github.com/jenkins-x/lighthouse/pkg/plugins"
+	"github.com/jenkins-x/lighthouse/pkg/repoowners"
 	fake2 "github.com/jenkins-x/lighthouse/pkg/scmprovider/fake"
 	"github.com/sirupsen/logrus"
+	"k8s.io/apimachinery/pkg/util/sets"
 )
 
+type fakeOwnersClient struct{}
```
ankitm123, line 34: ToDo: This should be in some test helper package as it's now being reused across tests in separate packages.

rawlingsj:
I'm don't think this is a good idea TBH, the aim for the automatic triggering of pipelines come from trusted users which should be managed by the organisation level permissions.  This was a key security design that Prow uses and deviating from that will introduce risk.  

I.e. if a person leaves a company their permissions are managed at the git provider organisation level.  Cleaning up OWNERS files requires updates to all related repos and things can be missed. 

It's important to distinguish between trusted users that can automatically trigger builds, and ultimately edit pipelines that get executed without review which is a significant security risk.  That is different from being in the OWNERS file as an approver of changes.

Let's leave the security model around trusted users and auto triggering pipelines as is.  This is a proven model and works across all Kubernetes repos, any change to this is a security risk.

rawlingsj:
This also means that anyone already in the OWNERS file could approve additional people that would be able to automatically trigger pipelines (and therefore change the behaviour), essentially bypassing the security model currently in place.

Closing this PR due to the security risks in introduces.

ankitm123:
> It's important to distinguish between trusted users that can automatically trigger builds, and ultimately edit pipelines that get executed without review which is a significant security risk

Thinking about it, this may make it less risky: If the user/bot is in OWNERS file and not in the github trusted org, then they cannot change the pipeline files (anything in .lighthouse folder basically) and OWNERS* file. We are detecting which files are being changed. Also this entire feature is turned off by default.

rawlingsj:
Someone could still write a unit test that would be executed without prior approval which did something malicious or change + approve code that should not be done once they have been removed from the org.  I think any change around the prow designed security model is not great TBH as we could open up for exploitation in a scenario like above or one that is not known yet.

ankitm123:
I think the specific use case for the issue addressed by this PR is dependency updates (go mod, package.json gem files etc ...), may be we only trigger for those PRs (so changes outside those files are not going to get triggered), so if a malicious actor opens a PR which changes code (for example tests) other than dependency updates, then it wont trigger (and we can add a comment in the PR explaning to the end user why it was not triggered).

Again by default, all of this is turned off.

The thing that worries me a bit is undesirable third party updates (for example, what happened to faker and color js).
It would be nice if certain bots/apps like renovate/action bot could be added to an org, but I guess that has more security risks ... Let me also ask around in prow channel to see what prow users think ...

m1pl:
After some brainstorming with @ankitm123 in Slack, this is what we gathered so far:
- Bot accounts are distinguishable by their `[bot]` suffix. No regular account can have that.
- It would be good to not hardcode a list of bots, but rather make it configurable (e.g. trustedBots array ["renovate", "dependabot"]).
- I'd rather not limit it to dependency files, maybe just blacklist OWNERS. But even this goes pretty far IMHO. The bots have access to the entire source code at least. If they're not trusted, they wouldn't be there in the first place. So artificially preventing users from doing stuff in jx that would otherwise be no problem is probably not the best idea. As long as we can distinguish bots from regular accounts, I don't see an issue. An organization admin needs to trust the bots, so they can open PRs.
- The PRs shouldn't need the `ok-to-test` label, other things like automerge, e.t.c. should be discussed separately if needed.

ankitm123:
For reference, this is the slack thread: https://kubernetes.slack.com/archives/C9MBGQJRH/p1639664112187200

ankitm123:
I think the source of the confusion was that I did not make it clear that the feature is for github apps (so let's only look at github for now), and not normal users. 
Also the code does not do such checks/distinction atm :facepalm: (the commit message also says users instead of github bots/apps).

m1pl:
Any progress here? This makes it a pain using jx with tools like Renovate or Dependabot.

joshuasimon-taulia:
I'm running self-hosted renovatebot using github app credentials and am unable to automatically test dependency PRs. I have tried configuring renovate to add the `updatebot` label to all PRs

```json
 "packageRules": [
    {
      "description": "Automatically merge minor and patch-level updates",
      "excludePackageNames": [
        "base",
        "gateway",
        "istiod"
      ],
      "matchUpdateTypes": [
        "minor",
        "patch",
        "digest"
      ],
      "addLabels": [
        "updatebot"
      ]
    }
  ]
```

PR #1421, opened 2022-02-16
fix: trigger jobs only when there is a GUID

### LINKED ISSUE #1420: Random duplicate jobs
Randomly, duplicate jx jobs are created a few seconds apart (2-3 seconds). 
This leads to wasting cloud resources.
Looking at the lighthouse webhook pod, only shows one lighthouse job was created, but build controller is creating 2 pipeline activities, which makes sense as there are 2 lighthouse jobs which creates 2 tekton pipelineruns.

I am not sure what is the root cause, but the second job does not have a GUID associated with it. It does have the `created-by-lighthouse` label set to `true`.
So, we can probably check if a GUID exists before creating a lighthousejob.

Long term: we want to understand the root cause of the duplicate jobs (a rogue goroutine may be)?

ankitm123:

fixes #1420

### DIFF
--- a/pkg/plugins/trigger/pull-request_test.go
+++ b/pkg/plugins/trigger/pull-request_test.go
@@ -17,6 +17,7 @@ limitations under the License.
 package trigger
 
 import (
+	"os"
 	"testing"
 
 	"github.com/jenkins-x/go-scm/scm"
@@ -259,7 +260,7 @@ func TestHandlePullRequest(t *testing.T) {
 	}
 	for _, tc := range testcases {
 		t.Logf("running scenario %q", tc.name)
-
+		os.Setenv("GIT_KIND", "github")
 		g := &fake2.SCMClient{
 			PullRequestComments: map[int][]*scm.Comment{},
 			OrgMembers:          map[string][]string{"org": {"t"}},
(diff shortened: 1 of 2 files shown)


pkg/plugins/trigger/pull-request.go
```diff
@@ -219,5 +220,8 @@ func buildAll(c Client, pr *scm.PullRequest, eventGUID string, elideSkippedConte
 	if err != nil {
 		return err
 	}
+	if eventGUID == "" && os.Getenv("GIT_KIND") == "github" {
```
ankitm123, line 223: Not sure how other scm providers deal with GUID, so let's keep it only for github

keskad:
What's the reason behind this?

ankitm123:
> What's the reason behind this?

Check the linked issue: #1420 
It's weird that there are no entries in the lighthouse pod logs. But the duplicate job was created by lighthouse. I am running this in my cluster for a week now, haven't seen a duplicate job since that.

keskad:
Damn, sorry I didn't notice the ticket. I see duplicated jobs on my test cluster, but I also see that github fires webhook twice for unknown for me reason.

ankitm123:
> but I also see that github fires webhook twice for unknown for me reason.

For the same event? I dont see that issue though, but wont be surprised if sometimes github does that.

keskad:
Yes. For the same event. Currently after bootjob passed and webhook was recreated I no longer see this issue. But previously I saw in github panel that github sent request twice in same second. Maybe that could be related somehow.

ankitm123:
/close
This actually will breal manually triggered pipelines!

PR #1425, opened 2022-02-25
feat: switch to ccache which is more maintained

ankitm123:

### DIFF
--- a/pkg/webhook/create_agent_integration_test.go
+++ b/pkg/webhook/create_agent_integration_test.go
@@ -4,7 +4,7 @@ import (
 	"os"
 	"testing"
 
-	lru "github.com/hashicorp/golang-lru"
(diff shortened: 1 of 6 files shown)


pkg/webhook/create_agent_integration_test.go
```diff
@@ -72,8 +72,7 @@ func TestCreateAgentIntegration(t *testing.T) {
 	s.FileBrowsers, err = filebrowser.NewFileBrowsers(serverURL, fb)
 	assert.NoError(t, err)
 
-	s.InRepoCache, err = lru.New(5000)
-	assert.NoError(t, err)
+	s.InRepoCache = lru.New(lru.Configure())
```
ankitm123, line 75: by default ccache sets the size to 5000.

PR #1426, opened 2022-03-01
fix: `patchDevImages.sh`  register arg not take effect

xinnjie:
I tried to change docker registery by add `-r` option, but it not take effect. After some debug, I change `if [ -n "${registry+}" ]` to `if [ -n "${registry}" ]`, then `-r` option take effect.

It looks like a typo

### DIFF
--- a/hack/patchDevImages.sh
+++ b/hack/patchDevImages.sh
@@ -68,7 +68,7 @@ for val in {0..5}; do
     image_names=(webhooks keeper foghorn gc-jobs tekton-controller jenkins-controller)
 
     full_image_name=""
-    if [ -n "${registry+}" ]
+    if [ -n "${registry}" ]
      then full_image_name="${full_image_name}${registry}/"
     fi
     if [ -n "${user}" ]


hack/patchDevImages.sh
```diff
@@ -68,7 +68,7 @@ for val in {0..5}; do
     image_names=(webhooks keeper foghorn gc-jobs tekton-controller jenkins-controller)
 
     full_image_name=""
-    if [ -n "${registry+}" ]
+    if [ -n "${registry}" ]
```
ankitm123, line 71: It should work as it's bash parameter expansion (`registry+` is a shorthand for `registry:+`) See: https://wiki.bash-hackers.org/syntax/pe
May be the error is somewhere else, let me check ...

xinnjie: If `registry` is set, then `${registry+}` would be empty, `if` statement would false, and `registry` would not add to `full_image_name`.
So I still think `+` is not needed.

PR #1437, opened 2022-03-17
fix: Page through statuses when polling

dippynark:
This PR ensures that Lighthouse poller finds all statuses when polling instead of just using the first page

### DIFF
--- a/pkg/poller/poller.go
+++ b/pkg/poller/poller.go
@@ -309,6 +309,9 @@ func (c *pollingController) ListAllStatuses(ctx context.Context, fullName string
 			return allStatuses, err
 		}
 		allStatuses = append(allStatuses, statuses...)
+		if response == nil {
+			break
+		}
 		// Check for an invalid value for the next page
 		if response.Page.Next <= page {
 			break
--- a/pkg/poller/poller.go
+++ b/pkg/poller/poller.go
@@ -300,11 +300,11 @@ func (c *pollingController) ListAllStatus(ctx context.Context, fullName string, 
 	allStatuses := []*scm.Status{}
 	page := 1
 	size := 100
-	opts := scm.ListOptions{
-		Page: page,
-		Size: size,
-	}
 	for {
+		opts := scm.ListOptions{
+			Page: page,
+			Size: size,
+		}
 		statuses, _, err := c.scmClient.Repositories.ListStatus(ctx, fullName, sha, opts)
 		if err != nil {
 			return allStatuses, err
--- a/pkg/poller/poller.go
+++ b/pkg/poller/poller.go
@@ -283,10 +283,7 @@ func (c *pollingController) pollPullRequestPushHook(ctx context.Context, l *logr
 }
 
 func (c *pollingController) hasStatusForSHA(ctx context.Context, l *logrus.Entry, fullName string, sha string, isRelease bool) (bool, error) {
-	opts := scm.ListOptions{
-		Page: 1,
-	}
-	statuses, _, err := c.scmClient.Repositories.ListStatus(ctx, fullName, sha, opts)
+	statuses, err := c.ListAllStatus(ctx, fullName, sha)
 	if err != nil {
 		return false, errors.Wrapf(err, "failed to list status")
 	}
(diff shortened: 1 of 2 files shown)


pkg/poller/poller.go
```diff
@@ -299,6 +296,28 @@ func (c *pollingController) hasStatusForSHA(ctx context.Context, l *logrus.Entry
...
+		}
+		allStatuses = append(allStatuses, statuses...)
+		// Check for an invalid value for the next page
+		if response == nil || response.Page.Next <= page {
+			break
+		}
+		page = response.Page.Next
+	}
+	return allStatuses, nil
+}
```
tomhobson, line 319: Can we get a test for this function :+1:

tomhobson (changes requested):

pkg/poller/poller.go
```diff
@@ -283,10 +283,7 @@ func (c *pollingController) pollPullRequestPushHook(ctx context.Context, l *logr
 }
 
 func (c *pollingController) hasStatusForSHA(ctx context.Context, l *logrus.Entry, fullName string, sha string, isRelease bool) (bool, error) {
-	opts := scm.ListOptions{
-		Page: 1,
-	}
-	statuses, _, err := c.scmClient.Repositories.ListStatus(ctx, fullName, sha, opts)
+	statuses, err := c.ListAllStatuses(ctx, fullName, sha)
```
ankitm123, line 286: can we move this function to go-scm instead?

dippynark: @ankitm123 I've tried to introduce a PR related to this to go-scm but it doesn't seem to be as active as this repository: https://github.com/jenkins-x/go-scm/pull/321

Either way, my understanding was that go-scm is providing a library that abstracts away from the APIs offered by different SCM providers, but here we are combining multiple API calls to get more functionality. Happy to close this PR though and move the logic into go-scm (although I believe this would require adding a new function to every supported provider), but if you agree with the above I can add the test that @tomhobson asked for

dippynark: I've created an initially PR: https://github.com/jenkins-x/go-scm/pull/323

If this looks like what you were thinking then I can add tests and close out this PR

ankitm123: I am back from vacation, and will review this PR tonight. Sorry for the delay in getting to this PR.

ankitm123: So, what I was thinking is to define another function `c.scmClient.Repositories.ListStatusAll(ctx, fullName, sha, opts)` in go-scm (checked ur go-scm repo). Instead of defining it in the client file, define it for each driver. I know some scm providers have different pagination limits (I recall having that issue with bitbucket) - that way we can easily extend the tests in go-scm.

But giving it a bit more thought:
I am fine with it being here, as u mentioned before, this is aggregation, and should be ok outside of go-scm. You can keep it here, and add the tests, and that will be ok for me.

ankitm123: >but it was implemented in exactly the same way for each provider, so tried to just create a general function.

The only concern I had was with pagination size, but I think it should be ok.

Cool, looking forward to the tests yay

dippynark: I think the pagination stuff gets handled by the providers as far as I know like this: https://github.com/jenkins-x/go-scm/blob/main/scm/driver/bitbucket/util.go#L35-L37

Have added a test, but it's not the best because the fake provider doesn't consider the ListOptions and always returns all the statuses, but I think it's still okay (or not sure how to do better without changes to go-scm): https://github.com/jenkins-x/go-scm/blob/main/scm/driver/fake/repo.go#L145-L150

dippynark: Sorry for linking across repositories but this PR is also quite important for us as otherwise Lighthouse poller has a race condition by potentially hiding previous statuses with newer ones: https://github.com/jenkins-x/go-scm/pull/321

ankitm123: > Have added a test, but it's not the best because the fake provider doesn't consider the ListOptions

That should be ok, we can update the fake provider, and update the tests in a subsequent PR...

ankitm123:
Can u rebase with the latest master? Else ur tests wont start (because of a recent tekton upgrade)

pkg/poller/poller.go
```diff
@@ -17,6 +17,7 @@ import (
 
 var (
 	censor = func(content []byte) []byte { return content }
+	StatusesPageSize = 100
```
ankitm123, line 20: should this be a const instead?

ankitm123 (approved):

PR #1444, opened 2022-05-06
feat: add CodeQL

aabhas-sao:


aabhas-sao:
a PR for CodeQL already exists.

PR #1463, opened 2022-07-16
fix: evaluate ignoreChanges before checking if pre-submit should always run

ankitm123:

ignore changes are not ignored if always_run is set to true.
So even if the user wants to not trigger a build when README file is changed for example, lighthouse will still trigger a build because always_run is evaluated first in the logic.

Context: https://kubernetes.slack.com/archives/C9MBGQJRH/p1657794048005119

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -36,31 +36,31 @@ build: build-webhooks build-poller build-keeper build-foghorn build-tekton-contr
 
 .PHONY: build-webhooks
 build-webhooks: ## Build the webhooks controller binary for the native OS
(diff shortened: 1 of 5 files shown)


Makefile
```diff
@@ -36,31 +36,31 @@ build: build-webhooks build-poller build-keeper build-foghorn build-tekton-contr
 
 .PHONY: build-webhooks
 build-webhooks: ## Build the webhooks controller binary for the native OS
-	$(GO) build -i -ldflags "$(GO_LDFLAGS)" -o bin/$(WEBHOOKS_EXECUTABLE) $(WEBHOOKS_MAIN_SRC_FILE)
```
ankitm123, line 39: `-i` is deprecated in go

ankitm123:
/hold 
going to add a test case for this, so we dont regress in the future.

rajatgupta24 (approved):
lgtm

PR #1467, opened 2022-07-27
feat: Sparse checkout

msvticket:
Sparse checkout is implemented especially for handling webhooks.

I have also changed to shallow clones even if you don't opt in to sparse checkout.

For now sparse checkout is enabled by setting the environment variable SPARSE_CHECKOUT=true.

I'm in the process of testing this in our own dev cluster now.

This is a part of solving jenkins-x/jx#8240

### DIFF
--- a/pkg/filebrowser/fake/fake_file_browser.go
+++ b/pkg/filebrowser/fake/fake_file_browser.go
@@ -63,7 +63,7 @@ func (f *fakeFileBrowser) getPath(owner, repo, path, ref string) string {
 	return filepath.Join(f.dir, path)
 }
 
-func (f *fakeFileBrowser) WithDir(owner, repo, ref string, fc filebrowser.FetchCache, fn func(dir string) error) error {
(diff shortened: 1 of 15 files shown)


msvticket:
If you like to test it you can use the image built from the PR. So these values needs to be set for the lighthouse chart:
```
env:
  SPARSE_CHECKOUT: true
webhooks:
  image:
    tag: 0.0.0-pr-1467-6-snapshot
```
To better see what is going on I also recommend setting the env `LOG_LEVEL: debug`

tomhobson:
I think we should default this -- it's a major perf improvement and we'll just make extra sure we're testing it properly

WDYT? @msvticket

msvticket:
As this PR is now: no. There are sometimes problems when reusing a previously checked out repo. I'll try to find time looking further into this. Maybe cutting this Gordian knot would be to always do a fresh clone when using sparse checkouts.

Also it's only tested with github.

tomhobson (approved):

PR #1470, opened 2022-08-07
only Issue owner and Issue Assignee can label hold

joker00777:
To not allow everyone to use hold label only the pr owner and the pr assignee can use them.


joker00777:
@ankitm123 just wanted to take your ideas if this looks cool than i will write the test case : )

ankitm123:
Not sure abt this, I think we want to allow all members of an org to be able to add/remove hold.

joker00777:
okai 
so should  i close it ?

PR #1472, opened 2022-08-25
fix: Reuse sparse checkout

msvticket:
It seems like --depth-1 is a bit to aggressive to work properly. Also it implies --single-branch, which means that checking out another branch might fail. To mitigate this I add --no-single-branch


msvticket:
I'm running lighthouse-webhook with this fix in our dev cluster. Since my last update yesterday I don't see any errors, while 154 lighthouse jobs have been created. This certainly an improvement. If you want to test this you can add the value to the lighthouse chart. I think this a crucial fix, so if nobody *opposes* I will merge this soon.

```
webhooks:
  image:
    tag: 0.0.0-PR-1472-2-SNAPSHOT
```

PR #1474, opened 2022-09-18
feat: Add periodic job support

dippynark:
This PR adds the scaffolding for a controller that implements the periodic jobs defined in the Lighthouse config. This is done by watching the Lighthouse ConfigMap to determine the list of periodic jobs to reconcile as described more in the [README](https://github.com/dippynark/lighthouse/blob/add-periodics-controller/cmd/strobe/README.md).

I have just done basic scaffolding within this PR to get an idea of what an implementation could look and to gather feedback for whether this is the right way to go. The actual reconciliation logic currently does nothing but could be implemented by either [templating and managing CronJobs](https://github.com/jenkins-x/lighthouse/issues/989) which periodically launch the actual LighthouseJobs or by creating a similar implementation to the CronJob controller (this is my preference): https://github.com/kubernetes-sigs/kubebuilder/blob/0b62a3ed0b7e3af53f013db4d9a31d96c7fdb7dd/docs/book/src/cronjob-tutorial/testdata/project/controllers/cronjob_controller.go

I stole the `strobe` name from [this previous implementation](https://github.com/jenkins-x/lighthouse/pull/992). An advantage of this implementation however is that watching for changes to the Lighthouse ConfigMap is event driven.

**UPDATE**:

A full implementation has now been added to this PR including Docker image build and Helm chart changes

A future improvement could be to use a LighthouseJob informer to reduce the number of list requests against the API server.


ankitm123:
It would also be good to look at how prow does this, after all lighthouse is a fork of prow.

dippynark:
@ankitm123 Hologram seems to work by checking every minute whether any periodic ProwJobs need running and then scheduling them accordingly:

- https://github.com/kubernetes/test-infra/blob/2dcd0ab8acd8018b03f93278d4de67e26b36fce6/prow/cmd/horologium/main.go#L134-L140
- https://github.com/kubernetes/test-infra/blob/2dcd0ab8acd8018b03f93278d4de67e26b36fce6/prow/config/config.go#L921-L926

This implementation works slightly differently by handling each periodic job individually. For example, it can individually retry a particular job with backoff if there is an issue and it will scale better as the number of the jobs increases (although for most users of Lighthouse I wouldn't expect these performance improvements to be noticeable).

I have now added a full implementation to this PR with few more details of the implementation documented.

ankitm123:
Need some time to review it, will complete the review this week.
Thank for this @dippynark , this is very nice! We can most likely replace github actions for some of the periodic jobs in jx org to test this out even more :100:

cmd/strobe/README.md
```diff
@@ -0,0 +1,27 @@
+# Strobe
+
+Strobe is a controller that implements the periodic jobs defined in the
+Lighthouse config ConfigMap:
+
+```yaml
```
ankitm123, line 6: I am not sure if it should be defined in a configmap. Normally in jx, configmap would be defined/configured in the cluster git repo, which should be normally off limits for repository owners. It should be in the same repo, so if a repository author wants to define a periodics, they should be able to define a new periodics job in the triggers yaml, wdyt?
May be a block like pre-submit/post-submit?

dippynark: Also sorry to reference other repositories, but would it be possible to get a review for this PR please? https://github.com/jenkins-x/go-scm/pull/321
We depend on the poller component being able to see if a particular job has ever succeeded (rather than just the latest run) which is the reason for the change. We are using a custom fork with that PR at the moment

dippynark: Thanks for the approval! I have created another PR to bump the go-scm version used by Lighthouse: https://github.com/jenkins-x/lighthouse/pull/1493

dippynark:
Hey @ankitm123, do you think any other changes are needed for this PR? I couldn't see how to get the TriggerConfig suggestion to work: https://github.com/jenkins-x/lighthouse/pull/1474#discussion_r989775410

Maybe it'd be a larger piece of work to get it working which I could do in a subsequent PR?

tomhobson:
@dippynark would you like to continue with this? I'd be happy to work with you to get this merged. It would be really cool to have!

cmd/strobe/README.md
tomhobson: I suppose the periodics would be registered similar to how pipelines are cached (but periodics will store instead). It being registered within the cluster repo goes away from the jenkins-x ideaollogy of being repo and developer centric as it would make you have to jump through 2 repos to add a feature to your repo

dippynark:
@tomhobson Yes this would be great! Sorry for the slow reply, haven't done much OSS stuff recently but would love to work more on this feature

Reading your comments it sounds like the next step would be to store the configuration for periodics whenever a webhook event is received (possibly this would require creating a ConfigMap per repository containing the periodic configuration for that repository, but perhaps a custom resource would be better for this)

What do you think about first getting this PR merged where you can only configure periodics through a ConfigMap like in the README: https://github.com/jenkins-x/lighthouse/blob/e4adcf7d2c513350dc3fc2ca6fc3efc3b192a184/cmd/strobe/README.md

And then subsequent PR(s) to add the support in the webhook for storing periodic configuration in the local cluster

msvticket:
This has been superseded by  #1563

/close

PR #1489, opened 2022-09-30
feat: adding pod labels in webhooks

joker00777:
Hi @ankitm123 we have a use case where we want to add some pod Labels : ). So it would be great if this can be done?


charts/lighthouse/values.yaml
```diff
@@ -125,6 +125,7 @@ webhooks:
 
   # webhooks.labels -- allow optional labels to be added to the webhook deployment
   labels: {}
+  podLables: {}
```
msvticket, line 128:
```suggestion
  podLabels: {}
```

charts/lighthouse/templates/webhooks-deployment.yaml
```diff
@@ -19,6 +19,9 @@ spec:
     metadata:
       labels:
         app: {{ template "webhooks.name" . }}
+{{- if  .Values.webhooks.podLables }}
+{{ toYaml .Values.webhooks.podLables | indent 8 }}
+{{- end }}
```
msvticket, line 24:
```suggestion
{{- if  .Values.webhooks.podLabels }}
{{ toYaml .Values.webhooks.podLabels | indent 8 }}
{{- end }}
```

charts/lighthouse/values.yaml
msvticket, line 128:
```suggestion
  podLabels: {}
```

msvticket (changes requested):
Please fix spelling of labels!

joker00777:
Done @msvticket : ).

msvticket (approved):

PR #1491, opened 2022-10-04
fix: Correct / update docs

msvticket:
Remove config for heart since there is no such plugin
Fix and run: make docs crd-manifests


crds/lighthouse.jenkins.io_lighthousejobs.yaml
```diff
@@ -2363,8 +2566,6 @@ spec:
                                                   type: object
                                               type: object
                                             type: array
-                                        required:
```
ankitm123, line 2366: Seems like the only thing that is removed from the card, probably we don't have a required marker :shrug:

ankitm123 (approved):
LGTM, the souce being removed as required is the only thing I am not sure about.

ankitm123:
Seems like it needs to be rebased with latest master, then keeper can merge it. Not sure why lighthouse has this.

PR #1503, opened 2022-10-12
fix: Make release work

msvticket:


ankitm123:
I was trying to fix it in this pr: https://github.com/jenkins-x/go-scm/pull/354

So whenever I build the updatebot image locally, and use that image it works, but when I use the one built in the release pipeline of jx-updatebot, it fails ... I am not sure why. The code is the same 🤔 

May be just revert back to non-alpine image? Wdyt?

msvticket:
> 
> So whenever I build the updatebot image locally, and use that image it works, but when I use the one built in the release pipeline of jx-updatebot, it fails ... I am not sure why. The code is the same 🤔

The Dockerfile uses the tag latest. Maybe you have another `latest` on your machine.

msvticket:
> May be just revert back to non-alpine image? Wdyt?

To answer that I need the answer to this question: What is the purpose of the jx-go image?

ankitm123:
> > So whenever I build the updatebot image locally, and use that image it works, but when I use the one built in the release pipeline of jx-updatebot, it fails ... I am not sure why. The code is the same 🤔
> 
> The Dockerfile uses the tag latest. Maybe you have another `latest` on your machine.

Thinking about it, that sounds like it could be the issue. I will delete my cache and check.

ankitm123:
> 

No idea honestly.

ankitm123:
> > May be just revert back to non-alpine image? Wdyt?
> 
> To answer that I need the answer to this question: What is the purpose of the jx-go image?

No idea honestly

PR #1511, opened 2022-11-07
WIP: feat: Add support for GIT_TOKEN_PATH environment variable

dippynark:
This PR adds support from retrieving the Git token from the filesystem instead of from an environment variable. In our case, we require this since retrieving secrets from environment variables is considered insecure since it is a common attack vector.

### DIFF
--- a/pkg/util/scmclient_test.go
+++ b/pkg/util/scmclient_test.go
@@ -1,0 +1,62 @@
+package util_test
+
+import (
+	"os"
+	"testing"
+	"path/filepath"
+
+	"github.com/jenkins-x/lighthouse/pkg/util"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestGetSCMToken(t *testing.T) {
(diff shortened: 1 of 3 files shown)


ankitm123:
I was away for the last 3-4 weeks, could u fix the merge conflicts?

dippynark:
@ankitm123 done!

pkg/util/scmclient_test.go
```diff
@@ -0,0 +1,65 @@
...
+	"path/filepath"
+	"testing"
+
+	"github.com/jenkins-x/lighthouse/pkg/util"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// This test ensures that we can retrieve the Git token from the environment
+func TestGetSCMToken(t *testing.T) {
```
ankitm123, line 14: NIT: It would be good to have table driven tests, but you can do that in a separate PR I suppose ...

ankitm123 (approved):

PR #1513, opened 2022-11-09
ci: switch to ko to build multiarch images

ankitm123:


ankitm123:
Once this works, we can add ko related pipeline config to go multi-arch catalog (atm, it uses cloud build)

PR #1531, opened 2023-03-23
feat: Add risk label

slimm609:
Add a `risk` label to allow assigning an impact label to merge requests. This provides an easy view of the overall impact of a change request and can faciliate the risk of multiple low vs high risk changes into a single code release.

### DIFF
--- a/pkg/plugins/label/label.go
+++ b/pkg/plugins/label/label.go
@@ -40,7 +40,7 @@ var (
 		ConfigHelpProvider: configHelp,
 		Commands: []plugins.Command{{
 			Prefix: "remove-",
-			Name:   "area|committee|kind|language|priority|sig|team|triage|wg|label",
+			Name:   "area|committee|kind|language|priority|sig|team|triage|wg|label|risk",
 			Arg: &plugins.CommandArg{
 				Pattern: ".*",
 			},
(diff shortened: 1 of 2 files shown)


pkg/plugins/label/label.go
```diff
@@ -40,7 +40,7 @@ var (
 		ConfigHelpProvider: configHelp,
 		Commands: []plugins.Command{{
 			Prefix: "remove-",
-			Name:   "area|committee|kind|language|priority|sig|team|triage|wg|label",
+			Name:   "area|committee|kind|language|priority|sig|team|triage|wg|label|risk",
```
tomhobson, line 43: Do you think risk should be a constrained label rather than a wildcard label?

Risk having a completely open ended label would limit extra features being built off of this in future

tomhobson (changes requested):
Let's discuss this change then we can go from there

pkg/plugins/label/label.go
slimm609: It would make it so it would have to be `/label risk/low` which is a little more inconvenient then `/risk low` and the other issue is in gitlab `/label` is reserved by gitlab itself so `/label` never makes it here

slimm609: is see `risk` similar to `priority` which is already in the list.

tomhobson (approved):
happy to move forward

PR #1556, opened 2023-06-28
feat: revert removal of lighthouse common pipelines & fix Tekton validation errors

Skisocks:
This PR reverts the changes made to remove lighthouse's Tekton pipeline parsing & fixes the issues seen with Tekton validating pipeline results values.

Issues with Tekton validation were due to new validation being added to the Pipeline.Results.Value object that checks the value is populated and correctly references a task within the pipelineSpec.

Added an info message that gets commented on PRs when opened if the pipelines in that repo use the `uses:` syntax.


Skisocks:
/hold

Holding for multiple reviews

msvticket:
> Issues with Tekton validation were due to new validation being added to the Pipeline.Results.Value object that checks the value is populated and correctly references a task within the pipelineSpec.

So it is this problem that is fixed?

```
  Warning  Failed            51m   PipelineRun  Pipeline jx/ga12-gke-gsm-dev-pr-1-verify-v4hdl can't be Run; it contains Tasks that don't exist: Couldn't retrieve Task "": error requesting remote resource: error getting "Git" "jx/git-0ed44de9af3fdfd292e37feaab97f83b": revision error: reference not found
  Warning  InternalError     51m   PipelineRun  1 error occurred:
           * Couldn't retrieve Task "": error requesting remote resource: error getting "Git" "jx/git-0ed44de9af3fdfd292e37feaab97f83b": revision error: reference not found

```

https://github.com/jenkins-x/jx3-versions/pull/3643#issuecomment-1597335473

Skisocks:
@msvticket No. That was a separate issue that was a result of going to native Tekton (a problem with the migration of pipeline-catalog). With reverting back to lighthouse's resolver then that error won't be relevant anymore. 

The problem that b7b8816bfa9565b718d8a41a8061fc0ec10acacb fixes is an issue internal to lighthouse that we noticed when we first upgraded Tekton within lighthouse. Basically Tekton is doing way more validation on things now and is checking that the PipelineResults.Value object exists and actually points to an existing task. Previously lighthouse didn't populate this, so the results object was completely broken.

msvticket (approved):

tomhobson (changes requested):
Can we get a test with both a git ref and a uses step to make sure that they _don't_ work together?

Skisocks:
Example message:
image

Skisocks:
Not sure about mentioning the issue though as it might get mentioned a tonne by jx pipelines.
image

msvticket:
> @tomhobson @msvticket Not sure about mentioning the issue though as it might get mentioned a tonne by jx pipelines. image

If you don't have access to the repo you won't see the message. (I don't see this.) And since most active repositories are bound to be private I don't think it's a problem as such. But I don't think we should add the comment before we actually have a suggested fix.

tomhobson (approved):
Happy to move forward as I think the messaging can tee up the changes we're going to make quite nicely.


We have users that aren't in slack that don't know this will be a future change. it would be nice to let them know before it happens.

PR #1563, opened 2023-07-25
feat: trigger on deploy status and periodically

msvticket:


msvticket:
Tests require jenkins-x/go-scm#404 to be merged

tomhobson:
@msvticket how we getting on with this? Need an assist or anything?

msvticket:
> @msvticket how we getting on with this? Need an assist or anything?

I still have some unit tests to implement. Some tests for periodics are implemented, but none for deployment. Also I haven't done any manual tests yet. 

Some documentation would be generated automatically, but some more should be written I guess.

So you could work on tests and docs... But I'll get to it eventually.

msvticket:
I have now gotten so far that I have started testing this in my own cluster. Triggering on deployment works. Here is a  screen shot from my pipeline dashboard showing a pipelinerun for a release and then a deployment:
screen shot from my pipeline dashboard showing a pipelinerun for a release and then a deployment

As for periodic triggers the scan and initialization for existing periodic triggers does not work; I don't think that is a merge stopper though since I put that behind a feature toggle. I will now test that adding a periodic trigger works.

tomhobson:
a m a z i n g

been waiting for this for so long, never had time to get it done

PR #1570, opened 2023-09-21
Trigger command

rochana-atapattu:
- pass trigger commands arg as a PipelineRunParam TRIGGER_COMMAND_ARG to presubmit
   with this change we can pass a parameter to out pipeline and use that parameter in the pipeline
   eg: 
   `/deploy dev`
   TRIGGER_COMMAND_ARG=dev
- ability to add multiple trigger commands LH_CUSTOM_TRIGGER_COMMAND will accept a comma seperated list of trigger

### DIFF
--- a/docs/trigger/README.md
+++ b/docs/trigger/README.md
@@ -2,7 +2,7 @@
 
 You can enable use LH_CUSTOM_TRIGGER_COMMAND environment variable to add custom triggers. It takes in a comma seperated list of commands.
 
-When these commands are triggered using ChatOps Lighthouse will convert the command arg into TRIGGER_COMMAND_ARG enviroment variable which you can refer from the pipeline.
+When these commands are triggered using ChatOps, Lighthouse will store the command arg in TRIGGER_COMMAND_ARG PipelineRunPara which you can refer from the pipeline.
 
 eg:
 `
--- a/docs/trigger/README.md
+++ b/docs/trigger/README.md
@@ -1,0 +1,44 @@
+# Lighthouse custome triggers
+
+You can enable use LH_CUSTOM_TRIGGER_COMMAND environment variable to add custom triggers. It takes in a comma seperated list of commands.
+
+When these commands are triggered using ChatOps Lighthouse will convert the command arg into TRIGGER_COMMAND_ARG enviroment variable which you can refer from the pipeline.
+
+eg:
+`
+LH_CUSTOM_TRIGGER_COMMAND=deploy,trigger
+`
+```
+apiVersion: config.lighthouse.jenkins-x.io/v1alpha1
+kind: TriggerConfig
+spec:
+  presubmits:
+  - name: pr
+    context: "pr"
+    always_run: true
+    optional: false
+    source: "pullrequest.yaml"
+  - name: trigger
+    context: "trigger"
+    always_run: false
+    optional: false
+    rerun_command: /trigger this
+    trigger: (?m)^/trigger( all| this),?(\s+|$)
+    source: "trigger.yaml"
+  - name: deploy
+    context: "deployr"
+    always_run: false
+    optional: false
+    rerun_command: /deploy dev 
+    trigger: (?m)^/deploy(?:[ \t]+([-\w]+(?:,[-\w]+)*))?(?:[ \t]+([-\w]+(?:,[-\w]+)*))?
+    source: "deploy.yaml"
+  postsubmits:
+  - name: release
+    context: "release"
+    source: "release.yaml"
+    branches:
+    - ^main$
+```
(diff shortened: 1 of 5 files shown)


msvticket:
Please give a description with the purpose of the PR

rochana-atapattu:
> Please give a description with the purpose of the PR

thank you for having a look at this, I have updated the description above

docs/trigger/README.md
msvticket, line 1: Is this really correct? Looking at the code it seem to me that documentation for the corresponding helm value is correct:

https://github.com/jenkins-x/lighthouse/blob/f5e770c38b351305beb918c851eb8e35eeda7d74/charts/lighthouse/values.yaml#L217-L219

pkg/plugins/trigger/generic-comment.go
```diff
@@ -94,6 +98,14 @@ func handleGenericComment(c Client, trigger *plugins.Trigger, gc scmprovider.Gen
...
 		return err
 	}
+  if arg != "" {
+    for i := range toTest {
+      toTest[i].Base.PipelineRunParams = append(toTest[i].Base.PipelineRunParams , job.PipelineRunParam{
+        Name: "TRIGGER_COMMAND_ARG",
+        ValueTemplate: arg,
+      })
+    }
+  }
```
msvticket, line 108: The feature should be documented and unit test added

msvticket (changes requested):
Your new code lines need to have their indentation adjusted to correspond to the surrounding code

docs/trigger/README.md
rochana-atapattu: even though it is documentation implies that we can create multiple triggers we actually can't.
https://github.com/jenkins-x/lighthouse/blob/cbf115a9757d4acf2ac84c2eae5ccee11578bdd5/charts/lighthouse/templates/webhooks-deployment.yaml#L47-L48
according to trigger.go
we take the env var
https://github.com/jenkins-x/lighthouse/blob/cbf115a9757d4acf2ac84c2eae5ccee11578bdd5/pkg/plugins/trigger/trigger.go#L41
and create a single trigger in init function
https://github.com/jenkins-x/lighthouse/blob/cbf115a9757d4acf2ac84c2eae5ccee11578bdd5/pkg/plugins/trigger/trigger.go#L78-L96
this  readme add in the docs for the new feature 05f1257fdd75b22c4d81e892df9660d779e97662 where we can add more one trigger. I will have to also update the values file to reflect the change

rochana-atapattu: 8af71b6ecd541eaeca197cb23e59f05aeff313cd adds in new documentation

pkg/plugins/trigger/generic-comment.go
rochana-atapattu: 836cdca8f763e195c00a4dedf80fa5020f0ea109 adding in unit tests. please let me know if I have missed anything. I updated the code so that handleGenericComment take in the command arg.
documentation for this feature is in
https://github.com/jenkins-x/lighthouse/blob/1b480280281582ed322d7eaa2131617fd364624a/docs/trigger/README.md?plain=1#L5

docs/trigger/README.md
msvticket: You mean that the existing documentation implied that multiple new prefixes could be added separated with comma?

I think you have misunderstood the vocabulary. Each entry in postsubmits and presubmits is a trigger. LH_CUSTOM_TRIGGER_COMMAND only adds an alternative way of activating the presubmit triggers. Why do you want to make it possible to add more prefixes for activating the trigger?

pkg/plugins/trigger/trigger.go
```diff
@@ -76,23 +76,22 @@ var plugin = plugins.Plugin{
...
-			Action: plugins.
-				Invoke(handleGenericCommentEvent).
-				When(plugins.Action(scm.ActionCreate), plugins.IsPR(), plugins.IssueState("open")),
-		}
-		plugin.Commands = append(plugin.Commands, customCommand)
-	}
-
-	plugins.RegisterPlugin(pluginName, plugin)
+  customTriggerCommand := os.Getenv(customerTriggerCommandEnvVar)
+  for _, trigger := range strings.Split(customTriggerCommand, ","){
```
dippynark, line 80: Can you not already do this using a regex pattern like: `build|plan|apply`?

msvticket: Yeah, I agree. This is interpreted as a regexp, so I think you should revert this change and update the documentation and tests accordingly.

msvticket (changes requested):

msvticket:
It seems like this PR is abandoned

PR #1571, opened 2023-10-04
feat: add source to lighthouse api

rochana-atapattu:
this change passes the sourcepath for the triggerconfig down to lighthousespec. with this change we can facilitate addition of new agents (engines), using the value in source to trigger pipelines of their own.


msvticket:
> this change passes the sourcepath for the triggerconfig down to lighthousespec. with this change we can facilitate addition of new agents (engines), using the value in source to trigger pipelines of their own.

Can you describe in a bit more detail how you envision this to be used?

msvticket:
PR is adandoned by developer

Issue #886, opened 2020-08-04
Have a way to NOT test a PR right away


vbehar:
Hi,

we have a use-case where we'd like to NOT execute the tests on a PR at the PR creation: our use-case is the creation of dozens of PRs when we bump the version of our jx buildpacks, or our base Helm chart for example. In that case, we want to create a PR on all the repos using the buildpack or base Helm chart, but triggering the tests on all these repos (there can be dozens or hundreds of them) might overwhelm the jx cluster.

In that case, we'd like to create the PRs but not run the tests automatically, and let the repos maintainers choose when they want to run the tests, using the chatops commands. Note that all our repos are private, and we can't use a bot that is not in our org otherwise it won't have access to our repos to create the PRs.

Ideally, we'd like to be able to add a label when creating the PRs, such as `needs-ok-to-test`, and the presence of such a label will disable the automatic trigger of the lighthouse jobs. Then, commenting with `/ok-to-test` will remove the label, and trigger the tests.

What do you think? Is there a better way to solve our issue that matches our use-case?

Thanks!

abayer:
So I think if you set `alwaysRun: false` but leave it in the required contexts list, the context won't run automatically but still will be required. Then `/test foo` will run it on demand. But a broader approach could be interesting, yeah.

abayer:
Ah, and yeah, what you're looking for is creating a PR and saying "don't run the tests for this PR right away" - we can definitely do something like that, probably like that label approach you mentioned. Seems reasonable/easy enough.

abayer:
And fwiw, `needs-ok-to-test` won't work - that gets automatically removed from any PR from a trusted user on reopen, and is ignored on open anyway. We'd need a new label for this, but that should be pretty trivial. I'm thinking `waiting-for-ok-to-test` - we then special-case the open/reopen/changes to the PR cases to skip building if `waiting-for-ok-to-test` present, and have `/ok-to-test` also remove `waiting-for-ok-to-test` as well as `needs-ok-to-test`.

vbehar:
yes exactly, most of the time we want the tests to run automatically (when devs create PRs), but for specific use-case like this one, we'd like to have the ability to run the tests later, manually.

I tried the`needs-ok-to-test` label, but it didn't work, as you said the code validates the PR author first, and in our case the bot is part of the org, so it's always able to activate the automatic trigger of the tests.

[...]

PR #1593, opened 2024-01-31
feat: allow to skip draft prs unless ready for review or /ok-to-test is added

JordanGoasdoue:
related to https://github.com/jenkins-x/lighthouse/pull/1476 && https://github.com/jenkins-x/lighthouse/issues/886

:warning: Once this PR merged we will need to 
- wait for lighthouse-client update
- update this jx-gitops PR: https://github.com/jenkins-x-plugins/jx-gitops/pull/953

Context: 
- We can't add label when opening a PR so we can only Draft a PR to disable the trigger of pipelines in case we need massive batch opening of PRs in 100+ repositories for instance . 
- Then we can activate manually the pipelines by clicking on 'Ready for Review' or commenting `/ok-to-test` from a trusted member

How to enable this feature ?
Add `skip_draft_pr` feature on the Trigger Plug-in from you scheduler:
- setup `skip_draft_pr` in scheduler trigger like `only_org_members` to allow easy migration of multiple team-app repositories in one time (Would be automatically applied to all repositories using the scheduler with this trigger.skip_draft_pr set to true)
- jx gitops scheduler would need to be run then from your bootjob to edit the plugin configmap 

How it works ?
- trigger pipelines in a Draft PR if `skip_draft_pr` is `false` depending if trusted or not (by default)
- trigger pipelines in a Draft PR if `skip_draft_pr` is `true` with `ok-to-test` label
- doesn't trigger pipelines in a Draft PR if `skip_draft_pr` is `true` but without `ok-to-test` label

- trigger pipelines when you click on `ready for review` if trusted and you don't have the ok-to-test label to avoid triggering the same pipelines than in previous draft mode before clicking the ready for review button 
- doesn't trigger pipelines when you click on `convert to draft` even if you have the ok-to-test label to avoid triggering the same pipelines than in previous non draft mode before clicking the convert to draft button 

Why this feature ?
- keep costs low when developers don't want to run builds
- Interesting when we need to open multiple Draft PRs in repositories using Jx without wanting to trigger all the repositories pipelines in same time

:point_down: :point_down: :point_down: :point_down: 
I have commented each testing scenarios in `pkg/plugins/trigger/pull-request_test.go` with the following PR Actions:
- `prAction: scm.ActionReadyForReview` 
- `prAction: scm.ActionConvertedToDraft`

to explain the logic behind when we click the "Ready for Review" or "Convert to Draft" button. 
I wanted here to avoid useless rerun of pipelines between both these modes.

```
go test ./pkg/plugins/trigger/ 
ok      github.com/jenkins-x/lighthouse/pkg/plugins/trigger     0.028s
```

:point_up: :point_up: :point_up: :point_up:

### DIFF
--- a/pkg/plugins/trigger/pull-request.go
+++ b/pkg/plugins/trigger/pull-request.go
@@ -18,6 +18,8 @@ package trigger
 
 import (
 	"fmt"
+	"net/url"
+
 	"github.com/jenkins-x/go-scm/scm"
 	"github.com/jenkins-x/lighthouse/pkg/config"
 	"github.com/jenkins-x/lighthouse/pkg/config/job"
@@ -28,7 +30,6 @@ import (
 	"github.com/jenkins-x/lighthouse/pkg/scmprovider"
 	"github.com/pkg/errors"
 	"github.com/sirupsen/logrus"
-	"net/url"
 )
 
 func handlePR(c Client, trigger *plugins.Trigger, pr scm.PullRequestHook) error {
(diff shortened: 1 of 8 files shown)


pkg/plugins/trigger/pull-request.go
```diff
@@ -242,23 +250,66 @@ I understand the commands that are listed [here](https://jenkins-x.io/v3/develop
...
-func TrustedPullRequest(spc scmProviderClient, trigger *plugins.Trigger, author, org, repo string, num int, l []*scm.Label) ([]*scm.Label, bool, error) {
-	// First check if the author is a member of the org.
-	if orgMember, err := TrustedUser(spc, trigger, author, org, repo); err != nil {
-		return l, false, fmt.Errorf("error checking %s for trust: %v", author, err)
-	} else if orgMember {
-		return l, true, nil
+func welcomeMsgForDraftPR(spc scmProviderClient, trigger *plugins.Trigger, pr scm.PullRequest) error {
+	org, repo, _ := orgRepoAuthor(pr)
+
+	comment := "This is a draft PR with trigger.SkipDraftPR enabled on this repository\n"
```
tomhobson, line 256: Is this for debugging, should we update this comment? Feels quite techy for a developer that maybe isn't familiar with jenkins x. We could always link docs alongside a more non jenkins x'y comment. or can we add a label instead similar to the needs-ok-to-test

tomhobson (changes requested):

pkg/plugins/trigger/pull-request.go
JordanGoasdoue: Screenshot_20240205_134636_Chrome

^ This would be the complete message for a Trusted user.
If the user is untrusted, then the message would be a little bit different.

Did you comment on the first part "This is a draft PR with trigger.SkipDraftPR enabled on this repository\n" or the whole message?

tomhobson: That's fine, if we can fix the spacing of the message then I'm happy to merge :)

JordanGoasdoue: Ok it's the new proposed version :+1: 

Screenshot from 2024-02-06 16-13-42


Screenshot from 2024-02-06 16-13-52

tomhobson:
/ok-to-test
/lgtm

Thanks for the PR!

tomhobson (approved):

PR #1613, opened 2024-08-27
Feat allow pipelinerun rerun to have github status update too

JordanGoasdoue:
Allow Github Status Update when Reruning a PipelineRun from Tekton Dashboard


charts/lighthouse/values.yaml
```diff
@@ -275,6 +275,9 @@ tektoncontroller:
   # tektoncontroller.terminationGracePeriodSeconds -- Termination grace period for tekton controller pods
   terminationGracePeriodSeconds: 180
 
+  # tektoncontroller.rerunPrReconciler -- Enables the rerun pipelinerun reconciler to run and update rerun pipelinerun status to github
```
msvticket, line 278: The usage message is cryptic if you don't happen to now the internals. Please change to something more user friendly. Also add the information to the README of the chart.

cmd/tektoncontroller/main.go
```diff
@@ -30,6 +31,7 @@ func gatherOptions(fs *flag.FlagSet, args ...string) options {
 	fs.StringVar(&o.namespace, "namespace", "", "The namespace to listen in")
 	fs.StringVar(&o.dashboardURL, "dashboard-url", "", "The base URL for the Tekton Dashboard to link to for build reports")
 	fs.StringVar(&o.dashboardTemplate, "dashboard-template", "", "The template expression for generating the URL to the build report based on the PipelineRun parameters. If not specified defaults to $LIGHTHOUSE_DASHBOARD_TEMPLATE")
+	fs.BoolVar(&o.rerunPipelineRunReconciler, "rerun-pr-reconciler", false, "Enables the rerun pipelinerun reconciler to run and update rerun pipelinerun status to github")
```
msvticket, line 34: Same here; please change to something more user friendly.

msvticket (changes requested):

charts/lighthouse/values.yaml
```diff
@@ -275,6 +275,9 @@ tektoncontroller:
   # tektoncontroller.terminationGracePeriodSeconds -- Termination grace period for tekton controller pods
   terminationGracePeriodSeconds: 180
 
+  # tektoncontroller.enableRerunStatusUpdate -- Enable updating the rerun status of PipelineRuns on GitHub
```
msvticket, line 278: "rerun status"? Maybe you mean

```suggestion
  # tektoncontroller.enableRerunStatusUpdate -- Enable updating the status on GitHub when PipelineRuns are rerun
```

msvticket (changes requested):

charts/lighthouse/values.yaml
JordanGoasdoue: yes it's better explained this way thank you

it's edited

msvticket (changes requested):
Now I realised another thing: We do support more than GitHub. So instead of writing "on GitHub" I think something like "at the git provider" would be better

PR #1621, opened 2024-12-06
Chore upgrade tekton pipelines

JordanGoasdoue:


msvticket:
According to https://github.com/tektoncd/pipeline/blob/main/docs/deprecations.md#v1beta1-deprecation the v1beta1 versions will be supported by tekton for a couple of years. So we should be able to support v1 and v1beta1 at the same time for a while, making the transition easier for us and our users.

There seems to be code to convert from v1beta1 to v1. For example: https://github.com/tektoncd/pipeline/blob/main/pkg/apis/pipeline/v1beta1/pipelinerun_conversion.go

I think the easiest would be to use such methods in pkg/triggerconfig/inrepo/load_pipelinerun.go if apiVersion is tekton.dev/v1beta1 in the file.

JordanGoasdoue:
@msvticket thank you will check this

PR #1624, opened 2024-12-12
feat: adding support for securityContext in helm chart

msvticket:
also fixing documentation


charts/lighthouse/values.yaml
```diff
@@ -197,6 +197,12 @@ webhooks:
   # webhooks.tolerations -- [Tolerations](https://kubernetes.io/docs/concepts/scheduling-eviction/taint-and-toleration/) applied to the webhooks pods
   tolerations: []
 
+  # webhooks.securityContext -- [Security Context](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) applied to the webhooks pods
+  securityContext: { }
```
Skisocks, line 201: Looks like the other object parentheses in the values don't have a space in them. All the new ones you've added have a space in.

```suggestion
  securityContext: {}
```

Skisocks (changes requested):

charts/lighthouse/values.yaml
msvticket: Right. Fixed now.

Skisocks: Yeah I've seen jetbrains does this by default

Skisocks (approved):

PR #1646, opened 2025-03-24
chore(keeper): remove automatic rerun logic for Tekton PipelineRuns with race condition failures

JordanGoasdoue:
### Summary

This PR removes the logic that was automatically retrying Tekton `PipelineRuns` when they failed due to known race conditions (like missing tasks or pipelines).

### Why

Keeper’s job is to manage PR pools and merging, similar to Prow’s Tide. It shouldn’t handle low-level CI issues or recreate failed `PipelineRuns`. That responsibility belongs to Tekton or a separate controller.

We're also in the process of upgrading to the latest Tekton version, and during testing, we noticed that Keeper was recreating `PipelineRuns` unexpectedly — which could interfere with how Tekton handles retries now.

To avoid this conflict and keep Keeper focused, we're removing the rerun logic completely.

### What’s removed

- The `rerunPipelineRunsWithRaceConditionFailure()` function
- The `pipelinerun_monitor.go` file and its tests
- The related call in `Sync()`


msvticket (approved):
I've tried to see if this code has been triggered in a fairly heavily used cluster of ours. It seem like no. So removing this looks good to me.

PR #1680, opened 2026-04-02
chore: use tini as init process for keeper

GavinPJK:
### Changes
* Install `tini` in the keeper Dockerfile and use as init

### Context
Keeper spawns git subprocesses during batch picking (`git fetch` -> `git-upload-pack`).

When git exits, orphaned subprocesses get re-parented to `keeper` itself (PID 1). Keeper doesn't effectively reap them, so they pile up as zombies and may exhaust the PID limit of the underlying node.

`tini` acts as a lightweight init process to handle this.

### DIFF
--- a/docker/keeper/Dockerfile
+++ b/docker/keeper/Dockerfile
@@ -1,9 +1,9 @@
 FROM alpine:3.23
 
-RUN apk add --update --no-cache ca-certificates git \
+RUN apk add --update --no-cache ca-certificates git tini \
     && adduser -D -u 1000 jx
 
 USER 1000
 
 COPY ./bin/keeper /home/jx/
-ENTRYPOINT ["/home/jx/keeper"]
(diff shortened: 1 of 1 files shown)


GavinPJK:
/cc @Skisocks @msvticket @tomhobson

We've had a discussion about this and running a separate init process is the simplest fix. Alternative solutions would involve much larger changes (changing keeper to manage process reaping itself; refactoring the git module to avoid subprocess spawning altogether)

msvticket (changes requested):
Add to the commit message why the change is needed. Also change the type from chore to fix, since it is, well, a fix.

msvticket (commented):
Have you considered setting `shareProcessNamespace: true` in the pod instead?

See for example https://www.back2code.me/2020/02/zombie-processes-back-in-k8s/

GavinPJK:
Closing and implementing your suggestion here https://github.com/jenkins-x/lighthouse/pull/1681

PR #1681, opened 2026-04-10
fix(keeper,poller,webhooks): set shareProcessNamespace in pod

GavinPJK:
### Changes
* Set `shareProcessNamespace: true` in pod specs for keeper, poller and webhooks

### Context
The git module uses `exec.Command()` to execute `git` commands. The spawned `git` processes will sometimes create subprocesses during batch picking (e.g.: `git fetch` -> `git-upload-pack`).

When `git` exits, the subprocesses are re-parented to PID 1 of the container. This is currently the application itself, which doesn't implement reaping. The defunct subprocesses accumulate and can eventually exhaust the PID limit of the  underlying node. Setting `shareProcessNamespace: true` moves PID 1 responsibility to the pod's shared pause container, which does properly reap zombies across all containers in the pod.

Setting `shareProcessNamespace: true` moves PID 1 responsibility to the pod's shared pause container, which does properly reap zombies across all containers in the pod. Only `keeper`, `poller`, and `webhooks` are affected — `foghorn` and the engine controllers make no subprocess calls.

### DIFF
--- a/charts/lighthouse/templates/keeper-deployment.yaml
+++ b/charts/lighthouse/templates/keeper-deployment.yaml
@@ -31,6 +31,7 @@ spec:
     spec:
       serviceAccountName: {{ template "keeper.name" . }}
       terminationGracePeriodSeconds: {{ .Values.keeper.terminationGracePeriodSeconds }}
+      shareProcessNamespace: true
       containers:
       - name: {{ template "keeper.name" . }}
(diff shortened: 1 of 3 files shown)


GavinPJK:
/cc @msvticket @Skisocks

Thank you for the suggestion, this approach is simpler

msvticket (approved):

Skisocks (approved):

PR #1689, opened 2026-05-11
fix(keeper): resolve base SHA via GetRef when PR list omits it

yelhouti:
GitLab's MR list endpoint does not return diff_refs / base_sha, so pr.BaseRef.Target.OID can be empty when keeper iterates the pool. This caused dividePool to seed sp.sha with "", which then propagated into LighthouseJob.Spec.Refs.BaseSHA and produced malformed PULL_REFS (e.g. "main,<pr-num>:<sha>:<ref>" instead of "main:<sha>,<pr-num>:<sha>:<ref>"), failing step-git-merge.

Resolve the base SHA via GetRef("heads/"+branch) when missing — once per unique subpool (typically one per repo since most PRs target the same branch), not per PR.

### DIFF
--- a/pkg/keeper/keeper.go
+++ b/pkg/keeper/keeper.go
@@ -1571,6 +1571,13 @@ func (c *DefaultController) dividePool(pool map[string]PullRequest, pjs []v1alph
 		}
 		fn := poolKey(org, repo, branch)
 		if sps[fn] == nil {
+			if baseSHA == "" {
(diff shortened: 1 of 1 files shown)


msvticket (approved):

msvticket:
Something is apparently weird with the PR since lighthouse can't merge it and also the Merge button is inactive for the PR on GitHub. Try rebasing.

yelhouti:
@msvticket just rebased lets see

PR #1693, opened 2026-06-08
feat(webhooks): add ingressKind and HTTPRoute template

GavinPJK:
### Changes
* Add `webhooks-httproute.yaml` template and default values to support Gateway API  
* Add `webhooks.ingressKind` for switching between ingress and httproute

### Context

Gateway API uses a `HTTPRoute` resource as a replacement for `Ingress`.

The `webhooks.ingressKind` key allows users to switch between the two implementation branches while remaining backward-compatible.

`httproute` users must provide a valid `parentRef` to a separately-managed `Gateway` resource. TLS configuration is the responsibility of the Gateway, so is not required here.

Part of https://github.com/jenkins-x/jx/issues/8962

### DIFF
--- a/charts/lighthouse/values.yaml
+++ b/charts/lighthouse/values.yaml
@@ -222,6 +222,23 @@ webhooks:
       # webhooks.ingress.tls.secretName -- Specify webhooks ingress tls secretName
       secretName: ""
 
+  httproute:
+    # webhooks.httproute.enabled -- Enable webhooks httproute
+    enabled: false
+
+    # webhooks.httproute.annotations -- Webhooks httproute annotations
+    annotations: {}
+
+    # webhooks.httproute.parentRefs -- List of parent Gateways the HTTPRoute binds to
+    # TLS is configured on the parent Gateway listener, not the HTTPRoute
+    parentRefs: []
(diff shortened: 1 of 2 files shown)


Skisocks (dismissed):
-7000 lines smells like something has broken right?

charts/lighthouse/values.yaml
```diff
@@ -203,6 +203,9 @@ webhooks:
   # webhooks.containerSecurityContext -- [Security Context](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/) applied to the webhooks containers
   containerSecurityContext: {}
 
+  # webhooks.ingressKind -- The kind of ingress to use for webhooks, either `ingress` or `httproute`
+  ingressKind: ingress
```
msvticket: What is the point of introducing this value and the httproute.enabled and ingress.enabled?

GavinPJK: Good catch - this was a hangover from when I was re-using some values from `webhooks.ingress` as part of the HTTPRoute configuration. Fixed now.

msvticket (approved):

PR #1696, opened 2026-07-07
Fix/postsubmit run if changed push compare

JordanGoasdoue:
## Summary

Fixes false postsubmit triggers on push when jobs use `run_if_changed` or `ignore_changes`.

**Root cause:** on push webhooks, Lighthouse unioned every `commits[]` file list instead of the net diff of the push. After a merge-from-main, stale paths in intermediate commits could still match matchers even when the net change was empty or unrelated.

**Fix:** new trigger config `push_changed_files` lets repos opt into SCM compare (`before` → `after`). The resolved file list is shared across all postsubmits on the push (one SCM call max via lazy cache).

## Behaviour

| `push_changed_files` | Effect |
|----------------------|--------|
| `all_commits` (default) | Legacy: union of all `commits[]` paths |
| `compare` | Net diff via `CompareCommits(before, after)` |

`compare` fallback to `all_commits` (with warning) when:
- SCM driver returns `ErrNotSupported`
- webhook has no `before` SHA (or zero SHA on new branch)
- no SCM client is available

Transient compare API errors still fail hard (no silent fallback).

Applies to **postsubmit change matching** — both `run_if_changed` and `ignore_changes` use the same changed-files provider.

## Configuration

Per repo, in the trigger plugin config (same place as `skip_draft_pr`):

```yaml
triggers:
  - repos:
      - org/repo
    push_changed_files: compare
```

No change required for repos that keep the default.


msvticket (commented):
I find the name of the configuration confusing. Wouldn't something like `changed_files_detection` be better?

PR #1698, opened 2026-07-09
fix(repo-owners): include path itself in OWNERS walk

Skisocks:
Seen issues with OWNERS files in sub-directories where lighthouse doesn't honour the OWNERS for a set of changed files within a sub dir. Seems that some places are passing in file paths when the should be passing in directories meaning the filepath.Dir() was being called on the dir containing the OWNERS file and so calculating for the wrong directory.

This bug has been covered up by the fact that all uses of fake owners in test cases are using different fake owners and some don't behave in the same way as the actual implementation. I've refactored the tests for various plugins so that they use the same fake owners. This shows all the failing tests due to the dir/path mix up.

I've then fixed the path/dir bug by including the path itself in the OWNERS walk. This means that both cases are handled correctly if path or dir is passed in as we always run from the bottom of the path/dir.

Note:
It appears this bug has been here since day 1 (~6 years) so I'm guessing that I'm the only person using the sub dir OWNERS logic 😂

### DIFF
--- a/pkg/plugins/approve/approve_test.go
+++ b/pkg/plugins/approve/approve_test.go
@@ -20,7 +20,6 @@ import (
 	"fmt"
 	"net/url"
 	"os"
-	"path/filepath"
 	"reflect"
 	"strings"
 	"testing"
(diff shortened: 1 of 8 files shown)


Skisocks:
/cc @msvticket

Probably best to review commit by commit. I intentionally broke the existing tests with the first commit to show the drift between the fakes and the actual implementation.

msvticket (approved):

Skisocks:
/hold

Gonna rebase

